PR #4, opened 2015-05-08
Fixed download bug (creating directory bug)

WPMGPRoSToTeMa:
Read more here: https://github.com/ValveSoftware/halflife/issues/1425#issuecomment-45697889


dreamstalker:
This code is client-side and should never be executed on server during normal operation since files uploading is disabled on server-side. But thanks anyway.

PR #5, opened 2015-05-08
Host say format and localize bug fix

WPMGPRoSToTeMa:


rehlds/engine/host_cmd.cpp
```diff
@@ -2569,10 +2574,18 @@ void Host_Say(qboolean teamonly)
 
 		host_client = client;
 
+#ifdef REHLDS_FIXES
+		PF_MessageBegin_I(MSG_ONE, RegUserMsg("TextMsg", -1), NULL, &g_psv.edicts[j + 1]);
```
dreamstalker: It's not obvious which problem have you fixed there. Please add a comment for clarification.

WPMGPRoSToTeMa: Text can be unsafe (format %, localize #) therefore need send text as argument. Use TextMsg instead of SayText, because SayText in Half-Life don't support arguments.

dreamstalker:
Thanks, going to merge it

PR #10, opened 2015-05-10
Function Host_UpdateStats on Linux doesn't work correctly.

s1lentq:
Fixed the computation of cpuPercent

Made signature checks (bz2, wad3) look more beatiful


dreamstalker:
Thanks, going to merge it

PR #12, opened 2015-05-11
Unrolled tail recursion in RecursiveHullCheck

theAsmodai:
Unrolled tail recursion in RecursiveHullCheck

### DIFF
--- a/rehlds/engine/pmovetst.cpp
+++ b/rehlds/engine/pmovetst.cpp
@@ -28,6 +28,9 @@
 
 #include "precompiled.h"
 
+// 1/32 epsilon to keep floating point happy
+#define	DIST_EPSILON	(0.03125f)
+
 int g_contentsresult;
 hull_t box_hull_0;
 box_clipnodes_t box_clipnodes_0;
@@ -638,8 +641,9 @@ struct pmtrace_s *PM_TraceLineEx(float *start, float *end, int flags, int usehul
 	return &tr;
 }
 
+#ifndef REHLDS_OPT_PEDANTIC
 /* <6ef4a> ../engine/pmovetst.c:844 */
-qboolean PM_RecursiveHullCheck(hull_t *hull, int num, float p1f, float p2f, vec_t *p1, vec_t *p2, pmtrace_t *trace)
+qboolean PM_RecursiveHullCheck(hull_t *hull, int num, float p1f, float p2f, const vec_t *p1, const vec_t *p2, pmtrace_t *trace)
 {
 	qboolean retval;
 	dclipnode_t *node;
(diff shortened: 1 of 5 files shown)


dreamstalker:
Thank you for the PR!
I have been looking through it for half an hour and realized that there are a huge changes in the code of 2 complex functions (PM_RecursiveHullCheck and SV_RecursiveHullCheck) and these changes, since they are under REHLDS_FIXES define, are not covered by integration tests.

I remember how many time I spent fixing bugs in these kind of functions, so my point is that these optimized functions must be tested in some way, to ensure that they work properly.

I suggest to do the following:
1. Split your pull request into several PRs. Let's start with unrolling a tail recursion in PM_RecursiveHullCheck. (That's it, without SSE optimizations)
2. It's time to introduce a new define for optimizations that don't change the outcome of the code/function when they're applied. Not changing the outcome allows us to test optimized code by existing integration tests. I'm not sure about the name of define, but let it be REHLDS_OPT_PEDANTIC
3. Put the whole optimized PM_RecursiveHullCheck function under the REHLDS_OPT_PEDANTIC define. Leave original function under the #else branch
4. Make sure REHLDS_OPT_PEDANTIC define is applied for all build configurations
4a. In Visual Studio just add it under the 'Configuration' dialog
4b. For gradle build script, add it to the rehlds/build.gradle, line 125
5. Build rehlds with gradle or create pull request to run integration tests. Make sure they are passed.

I will also add some comments to your code

rehlds/engine/pmovetst.cpp
```diff
@@ -651,125 +653,141 @@ qboolean PM_RecursiveHullCheck(hull_t *hull, int num, float p1f, float p2f, vec_
...
-		midf = 0.0;
-	}
+		// find the point distances
+		node = &hull->clipnodes[num];
+		plane = &hull->planes[node->planenum];
+		if (plane->type >= 3u)
+		{
+#ifdef REHLDS_FIXES
+			t1 = _DotProduct(p1, plane->normal) - plane->dist;
+			t2 = _DotProduct(p2, plane->normal) - plane->dist;
```
dreamstalker: No need to put this code under REHLDS_FIXES
_DotProduct does the same and looks more read-friendly. Use of SSE's dot product is applied inside _DotProduct() and has its own REHLDS_FIXES define

rehlds/engine/sys_dll2.cpp
```diff
@@ -234,6 +237,48 @@ NOXREF void Sys_CheckOSVersion(void)
...
+#define SSSE3_FLAG (1<<9);
+#define SSE4_1_FLAG (1<<19);
+#define SSE4_2_FLAG (1<<20);
+#define AVX_FLAG (1<<28);
+#define AVX2_FLAG (1<<5);
+
+void Sys_CheckCpuInstructionsSupport(void)
+{
+	__asm
+	{
```
dreamstalker: Please avoid writing asm code if it's possible to use intrinsics.
Check sys_shared.cpp and move CPU feature detection there

rehlds/engine/world.cpp
```diff
@@ -1119,7 +1180,7 @@ trace_t SV_Move(const vec_t *start, const vec_t *mins, const vec_t *maxs, const
 		clip.end = worldEndPoint;
 		worldFraction = clip.trace.fraction;
 
-		clip.type = (unsigned char)type;
+		clip.type = type & 0xF;
```
dreamstalker: What have you fixed here?

PR #13, opened 2015-05-11
Math optimizations, faster traceline version

theAsmodai:
Added SSE 4.1 implementation of dot product, added SinCos function, maked fast version of SV_Move for point hull (trace line), small copying optimization in SV_RunCmd.


dreamstalker:
Great! Merging it.

PR #14, opened 2015-05-11
Maked reallocation of resource names in precache.

s1lentq:


rehlds/engine/pr_cmds.cpp
```diff
@@ -1124,8 +1128,11 @@ short unsigned int EV_Precache(int type, const char *psz)
 				char* evScript = (char*) COM_LoadFile(szpath, 5, &scriptSize);
 				if (!evScript)
 					Host_Error("EV_Precache:  file %s missing from server\n", psz);
-
+#ifdef REHLDS_FIXES
+				g_psv.event_precache[i].filename = ED_NewString(psz);
```
dreamstalker: Please add a comment in the code about the issue you have fixed

theAsmodai:
Ctrl+C, Ctrl+V
Many modders don't know that the resource names passed to precache functions must be a static strings. Also some metamod modules can be unloaded during the server running. Thereafter the pointers to resource names, precached by this modules becomes invalid and access to them will cause segfault error. Reallocating strings in the rehlds temporary hunk memory helps to avoid these problems.

s1lentq:
@theAsmodai Thanks.
@dreamstalker All done.

dreamstalker:
Thanks, merging it

PR #26, opened 2015-05-25
JITted DELTA_TestDelta.

theAsmodai:
Maked test for DELTA_TestDelta.
Markbits aligned to 16.
Unrolled tail recursion in SV_FindTouchedLeafs and SV_LinkContents.
Other small optimizations and fixes.

### DIFF
--- a/rehlds/engine/delta_jit.cpp
+++ b/rehlds/engine/delta_jit.cpp
@@ -27,7 +27,7 @@ unsigned int DELTAJIT_GetFieldSize(delta_description_t* desc) {
 		return 4;
 
 	case DT_STRING:
-		return desc->fieldSize;
+		return 0;
 
 	default:
 		rehlds_syserror("%s: Unknown delta field type %d", __FUNCTION__, desc->fieldType);
@@ -84,6 +84,7 @@ void DELTAJIT_CreateDescription(delta_t* delta, deltajitdata_t &jitdesc) {
 		jitField->offset = fieldDesc->fieldOffset;
 		jitField->type = fieldDesc->fieldType;
 		jitField->length = fieldSize;
+		jitField->significantBits = fieldDesc->significant_bits;
 
 		if (fieldDesc->fieldType != DT_STRING) {
 			bool firstBlock = true;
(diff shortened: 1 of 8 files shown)


rehlds/engine/delta_jit.cpp
```diff
@@ -375,25 +379,17 @@ CDeltaClearMarkFieldsJIT::Result CDeltaClearMarkFieldsJIT::main(Addr src, Addr d
...
-		movd(xmm1, edx);
-		psllq(xmm0, 32);
-		por(xmm0, xmm1);
-		
-		movdqu(xmm_tmp, xmmword_ptr[eax]);
-		pand(xmm_tmp, xmm0); //clean high 64 bits
-
-		por(marked_fields_mask, xmm_tmp); //apply the 'force' mask
+		movdqu(xmm0, xmmword_ptr[eax]);
+		movq(xmm_tmp, xmm0); //clean garbage in high 64 bits of force mask
```
dreamstalker: Is it possible to use movq(xmm_tmp, qword_ptr[eax]) there?

rehlds/engine/delta_jit.cpp
```diff
@@ -452,15 +448,235 @@ void CDeltaClearMarkFieldsJIT::processStrings(Addr src, Addr dst) {
 	}
 }
 
-CDeltaJit::CDeltaJit(delta_t* _delta, CDeltaClearMarkFieldsJIT* _cleanMarkCheckFunc) {
+class CDeltaTestDeltaJIT : public jitasm::function<int, CDeltaTestDeltaJIT, void*, void*, void*>
```
dreamstalker: The code of this class is 90% copypasted from CDeltaClearMarkFieldsJIT, which is bad practice. Consider using inheritance instead - move logic that does iteration over all fields into parent class and implement 'on field changed' logic in child classes

rehlds/engine/delta_jit.cpp
```diff
@@ -375,25 +379,17 @@ CDeltaClearMarkFieldsJIT::Result CDeltaClearMarkFieldsJIT::main(Addr src, Addr d
...
 	EndIf();
 
 	size_t delta_markbits_offset = offsetof(CDeltaJit, marked_fields_mask);
 
 	//Save mask from SSE register to CDeltaJit::marked_fields_mask and CDeltaJit::originalMarkedFieldsMask
 	mov(ebx, ptr[deltaJit]);
-	movdqu(xmmword_ptr[ebx + delta_markbits_offset], marked_fields_mask);
-	movdqu(xmmword_ptr[ebx + offsetof(CDeltaJit, originalMarkedFieldsMask)], marked_fields_mask);
+	movdqa(xmmword_ptr[ebx + delta_markbits_offset], marked_fields_mask);
+	movdqa(xmmword_ptr[ebx + offsetof(CDeltaJit, originalMarkedFieldsMask)], marked_fields_mask);
```
dreamstalker: Is it possible to use movq there? In that case we may also get rid of high_sse_bits and alignment in CDeltaJit

rehlds/engine/delta_jit.cpp
theAsmodai: Yes, it works on x86.

PR #30, opened 2015-06-07
Fixed message fragments validation.

theAsmodai:
Fixed NET_GetLong.
Passing netadr_t to functions by reference instead of copying.
Added missed Q_ prefix for stdlib functions.
Added names for network constants.


dreamstalker:
Thx, merging it

PR #35, opened 2015-06-15
Voice_GetClientListening bugfix

WPMGPRoSToTeMa:


dreamstalker:
Could you please create an issue that was fixed by this change?
It would be much better for people who faced this bug to find a solution.

PR #43, opened 2015-06-30
Fix for the Buffer Overload 4 exploit

thecrock:
Added rate limiter for client's move commands (SV_ParseMove)
Default limits are 350cmds/sec average (1.5 sec interval) and 600 cmds/sec burst (0.2 sec interval)
Client running at 100fps and cl_cmdrate 100 with ideal network connection has 100cmds/seconds rate.
In case of bad network (50% packet loss) server would execute 200cmds/sec for this player. It's pretty hard to hit the 350cmds/sec limit in normal condition.

### DIFF
--- a/rehlds/engine/sv_user.cpp
+++ b/rehlds/engine/sv_user.cpp
@@ -1560,6 +1560,19 @@ void SV_ParseMove(client_t *pSenderClient)
 		net_drop = 0;
 	}
 
+	//check move commands rate for this player
+#ifdef REHLDS_FIXES
+	int numCmdsToIssue = numcmds;
+	if (net_drop > 0) {
+		numCmdsToIssue += net_drop;
+	}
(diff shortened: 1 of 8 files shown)


rehlds/engine/sv_user.cpp
```diff
@@ -1560,6 +1560,15 @@ void SV_ParseMove(client_t *pSenderClient)
 		net_drop = 0;
 	}
 
+	//check move commands rate for this player
+#ifdef REHLDS_FIXES
+	g_MoveCommandRateLimiter.MoveCommandsIssued(host_client - g_psvs.clients, net_drop + 1);
```
dreamstalker: Are you sure about the "net_drop + 1"?
I think numcmds should also be taken into account.

thecrock:
Increased limits and fixed counting of move commands

dreamstalker:
Looks good, merging it.

PR #45, opened 2015-07-04
Added client_t<>index conversion functions to ReHLDS API.

theAsmodai:


rehlds/rehlds/rehlds_interfaces_impl.cpp
```diff
@@ -136,6 +136,19 @@ IGameClient* EXT_FUNC CRehldsServerStatic::GetClient(int id)
...
+{
+	if (id < 0 || id >= g_psvs.maxclients)
+		Sys_Error(__FUNCTION__": invalid id provided: %d", id);
+
+	return &g_psvs.clients[id];
+}
+
+int EXT_FUNC CRehldsServerStatic::GetIndexOfClient_t(client_t* client)
+{
+	return int(client - g_psvs.clients);
```
dreamstalker: I believe vailadation of the client is required here: against NULLs, boundary of svs.clients and alignment in svs.clients (i.e. when the client argument points to somewhere between clients[0] and clients[1])

PR #52, opened 2015-07-12
Added SetModelName function to ReHLDS ServerData

WPMGPRoSToTeMa:


theAsmodai:
What could be the reason to setting fake mapname?

WPMGPRoSToTeMa:
Fake mapname need for FastDlFix metamod plugin

PR #57, opened 2015-07-15
Added PrepareUserInfoToTransmit callback to ReHLDS API

WPMGPRoSToTeMa:


rehlds/engine/sv_main.cpp
```diff
@@ -3713,6 +3713,8 @@ void SV_FullClientUpdate(client_t *cl, sizebuf_t *sb)
 	info[sizeof(info) - 1] = 0;
 	Info_RemovePrefixedKeys(info, '_');
 
+	g_RehldsHookchains.m_PrepareUserInfoToTransmit.callChain(nullptr, GetRehldsApiClient(cl), info);
```
dreamstalker: I think you should also pass the length of the info buffer to this chain, in order to manipulate that buffer safely.

theAsmodai:
I think would be useful add to hookchain 'receiver' argument for customizations. For example, that different players can see different models. I once did this for CSDM FFA.

WPMGPRoSToTeMa:
@theAsmodai then we need change following code to for loop:
```C++
{
	client->sendinfo = FALSE;
	client->sendinfo_time = realtime + 1.0;
	SV_ExtractFromUserinfo(client);
	SV_FullClientUpdate(client, &g_psv.reliable_datagram);
}
```
But what macro we need choose for code separation?

theAsmodai:
It's a little more complicated than I thought. The recipient can be determined by msg and host_client, but for customization we must also change the destination buffer. Maybe will be better to make a separate hookchain for the whole SV_FullClientUpdate function, and implement customization logic in metamod module.

theAsmodai:
@WPMGPRoSToTeMa, I created separate PR for it: https://github.com/dreamstalker/rehlds/pull/62

dreamstalker:
Obsoleted by #62

PR #59, opened 2015-07-16
ReHLDS API version print by "version" command

WPMGPRoSToTeMa:


rehlds/engine/host_cmd.cpp
```diff
@@ -2427,6 +2427,7 @@ void Host_Version_f(void)
 {
 	Con_Printf("Protocol version %i\nExe version %s (%s)\n", PROTOCOL_VERSION, gpszVersionString, gpszProductString);
 #ifdef REHLDS_FIXES
+	Con_Printf("ReHLDS API version %i.%i\n", REHLDS_API_VERSION_MAJOR, REHLDS_API_VERSION_MINOR);
```
dreamstalker: Checking API version may also be useful for pure builds, so it's better to move this outside REHLDS_FIXES block

PR #60, opened 2015-07-16
Fixed MAX_FRAGMENTS count

WPMGPRoSToTeMa:


rehlds/engine/net.h
```diff
@@ -250,7 +250,7 @@ typedef struct flow_s
 #ifndef REHLDS_FIXES
 #define MAX_FRAGMENTS 25000
 #else
-#define MAX_FRAGMENTS ((NET_MAX_PAYLOAD + FRAGMENT_SIZE - 1) / FRAGMENT_SIZE) // should be enough for any send buf
+#define MAX_FRAGMENTS (NET_MAX_PAYLOAD / FRAGMENT_SIZE) // should be enough for any send buf
```
dreamstalker: I don't get what you are intended to fix there, so please provide detailed description of an issue

WPMGPRoSToTeMa: Old variant: MAX_FRAGMENTS * FRAGMENT_SIZE > NET_MAX_PAYLOAD
New variant: MAX_FRAGMENTS * FRAGMENT_SIZE <= NET_MAX_PAYLOAD
If MAX_FRAGMENTS * FRAGMENT_SIZE > NET_MAX_PAYLOAD, then net_message in Netchan_CopyNormalFragments can overflow.

dreamstalker: In that case it's better to add an overflow check in Netchan_CopyNormalFragments

PR #61, opened 2015-07-16
Added checks for client resource list (customization list)

WPMGPRoSToTeMa:


rehlds/engine/sv_upld.cpp
```diff
@@ -408,6 +408,14 @@ void SV_ParseResourceList(client_t *pSenderClient)
 	SV_ClearResourceList(&host_client->resourcesneeded);
 	SV_ClearResourceList(&host_client->resourcesonhand);
 
+#ifdef REHLDS_CHECKS
```
dreamstalker: It's better to use REHLDS_FIXES here.
REHLDS_CHECKS is used for checks the violation of which will/may result in a crash (mostly for buffer overflows)

PR #62, opened 2015-07-16
Implemented SV_WriteFullClientUpdate callback

theAsmodai:
It's my vision of callback that allows to customize sending client's userinfo. But there is a problem, if info is sent to all, for customization we must call SV_WriteFullClientUpdate individually for each client. What's better:
1) Rewrite callNext to allow call it several times from one chain.
2) Write another call next function for this situation.
3) Add api function for SV_WriteFullClientUpdate and use it instead of call next chain.

### DIFF
--- a/rehlds/engine/sv_main.cpp
+++ b/rehlds/engine/sv_main.cpp
@@ -3713,28 +3713,33 @@ int SV_CalcPing(client_t *cl)
 	return 0;
 }
 
-/* <a874a> ../engine/sv_main.c:4991 */
-void SV_FullClientUpdate(client_t *cl, sizebuf_t *sb)
+void EXT_FUNC SV_WriteFullClientUpdate_internal(IGameClient *client, char *info, size_t maxlen, sizebuf_t *sb, IGameClient *receiver)
 {
-	unsigned int i;
-	char info[MAX_INFO_STRING];
+	client_t* cl = client->GetClient();
 	MD5Context_t ctx;
 	unsigned char digest[16];
-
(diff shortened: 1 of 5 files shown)


rehlds/engine/server.h
```diff
@@ -688,7 +688,7 @@ void SV_ReadPackets(void);
 //NOBODY int htons(void);
 void SV_CheckTimeouts(void);
 int SV_CalcPing(client_t *cl);
-void SV_FullClientUpdate(client_t *cl, sizebuf_t *sb);
+void SV_FullClientUpdate(client_t *cl, sizebuf_t *sb, client_t *receiver);
```
dreamstalker: This breaks binary compatibility with original hlds in the 'hooker' mode, i.e. when rehlds installs hooks into original engine

theAsmodai: Forgot for it.

dreamstalker: We don't have a flag/define for changes that break functions- and structs- level compatibility with original hlds. Right now you can use 'hooker' mode even with REHLDS_FIXES turned on. It's better to rewrite this code in a way that won't change signatures of existing functions

theAsmodai: While yes, there is a simple solution.

theAsmodai:
What's about of several calls of chain->callNext(...) for each cl->netchan.message buffer instead of sv.reliable_datagram? For now it's will not working.

dreamstalker:
Created an issue #63

PR #81, opened 2015-10-17
Fixed loading of maps without world classname or with long list of wads

theAsmodai:
Removed rtti
Reduced MAX_HOOKS_IN_CHAIN value


latinas:
World classname is prefix of mapname, for example de_, cs_, as_, etc. I'm right?

theAsmodai:
No, in some maps the world entity defined without 'classname = worldspawn' parameter.

latinas:
I got it, thx

PR #82, opened 2015-10-17
Added optimized stdlib functions from Agner's asmlib

theAsmodai:


latinas:
Are this functions works faster?

theAsmodai:
Yes. For now rehlds uses default string and mem functions written using SSE2. It's old technology, but supported by most of cpus. Asmlib can use newest instruction extensions (up to AVX2) if it supported. Also, the library is written by one of the best optimization specialists.

latinas:
Great! Thanks for your answer

PR #87, opened 2015-10-27
Cut of nicknames '+' if next there is something digit or alphanumeric character.

s1lentq:
Added prefix Q_* for functions strlen/strcmp/strpcy/memset/memcpy/memmove etc.
Added for solution configuration (Debug Swds Play/Release Swds Play) the path to libacof32.lib in linker.

### DIFF
--- a/rehlds/engine/common.cpp
+++ b/rehlds/engine/common.cpp
@@ -34,8 +34,8 @@ char gpszVersionString[32];
 char gpszProductString[32];
 
 char* strcpy_safe(char* dst, char* src) {
-	int len = strlen(src);
-	memmove(dst, src, len + 1);
+	int len = Q_strlen(src);
+	Q_memmove(dst, src, len + 1);
 	return dst;
 }
 
--- a/rehlds/engine/common.h
+++ b/rehlds/engine/common.h
@@ -135,7 +135,6 @@ extern int loadsize;
 #define Q_strnicmp _strnicmp
 #define Q_strstr A_strstr
 #define Q_strchr strchr
-#define Q_strrchr strrchr
 #define Q_strlwr A_strtolower
 #define Q_sprintf sprintf
 #define Q_snprintf _snprintf
--- a/rehlds/engine/ed_strpool.cpp
+++ b/rehlds/engine/ed_strpool.cpp
@@ -3,12 +3,12 @@
 class CStringPoolMap : public CStaticMap<const char*, char*, 8, 2048> {
 protected:
 	virtual uint32 hash(const char* const &val) {
-		unsigned int len = strlen(val);
+		unsigned int len = Q_strlen(val);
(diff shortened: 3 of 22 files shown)


latinas:
I have one question: what feature in Q_ functions? They works faster or smthng else?

s1lentq:
@latinas AFAIR Asmodai already answered your question https://github.com/dreamstalker/rehlds/pull/82#issuecomment-148952932

latinas:
Ah, i thought this functions and functions Asmodai added in 82 commit are different. Ok, thanks for answer

theAsmodai:
Functions with Q_ prefix is only a definitions that makes it easy to replace standart library functions.

WPMGPRoSToTeMa:
Why do you need cut '+' characters of nicknames?

theAsmodai:
Since this is an unused feauture causing only problems.

WPMGPRoSToTeMa:
@theAsmodai what feature?

theAsmodai:
I don't know useful applications of displaying buttons with binds.

WPMGPRoSToTeMa:
@theAsmodai are you about ShowMenu message?

theAsmodai:
Yes.

WPMGPRoSToTeMa:
First of all ShowMenu message is a part of game, not engine. Therefore this should be fixed in game part (gamedll/clientdll), but not in engine. Secondly you can remove/replace this characters from nickname only for ShowMenu in plugins code. ('+' in scoreboard doesn't cause bugs)

theAsmodai:
The engine contains so many dirty hacks (especially for CS). Since there is no way to fix the client, we have to do so.

WPMGPRoSToTeMa:
@theAsmodai, but it can be fixed in gamedll.

theAsmodai:
How? Changing some chars will lead to mismatched nicknames in different places and collision problems.

WPMGPRoSToTeMa:
@theAsmodai you can replace '+' to U+2795 before show it in 'bugged' menu.

theAsmodai:
It's possible if nickname doesn't take all available length. But this correction must be done in a hundred places.

WPMGPRoSToTeMa:
@theAsmodai this is plugins problem, not engine and gamedll. Show nicknames in ShowMenu is '
undocumented feature'. Why the plugins problem should be fixed in engine?

theAsmodai:
Be realistic. Otherwise, the problem will never be solved.

WPMGPRoSToTeMa:
The ShowMenu problem should be fixed in ShowMenu. So you can hook ShowMenu message and replace/remove characters in it, but don't do this in nickname.

theAsmodai:
Too complicated for such a simple thing. The '+' char in names are very rarely used.

PR #100, opened 2015-12-04
Add: Cut character 0x1160

s1lentq:
use own tables non-printable characters for windows.
character: 0x1160 marked to like non-printable

this character is skipped
name "ᅠ"

More information about the symbol http://unicode-table.com/en/1160/


WPMGPRoSToTeMa:
From 0x1160 to 0x117F? What you can say about 0x115F?

s1lentq:
@WPMGPRoSToTeMa 

> From 0x1160 to 0x117F?

I was mistaken a little, now fixed

> What you can say about 0x115F?

maybe just changed 0x83FFFFFF -> 0x03FFFFFF ?

Issue #114, opened 2016-01-04
Sprays not working correctly.


IgnacioFDM:
Many (All?) custom sprays are not working.

Steps to reproduce:
1)Set the following example tempdecal.wad (https://www.sendspace.com/file/lfwbds) in your client cstrike directory.
2)Try to connect to ReHLDS server (make sure sv_allowuploads 1).

Client will get stuck in "Precaching resources..."

Same steps on regular HLDS result in client connecting properly and spray showing properly.

If sv_allowuploads is set to 0, then client will be able to connect (sprays won't show of course).

Also, related to this are segfault crashes after changelevel.

This line

https://github.com/dreamstalker/rehlds/blob/fef6405b6185b08c10d8e4585be2711063728856/rehlds/engine/hashpak.cpp#L366

Will result in segfault at random after changelevel (most likely related to what the clients' sprays are).
This gets called from 

```c
void HPAK_FlushHostQueue(void)
```

which gets called from
```c
void EXT_FUNC SV_ActivateServer_internal(int runPhysics)
```

(Github isn't accepting my zip file)
https://www.sendspace.com/file/lfwbds

PR #134, opened 2016-01-17
Fixed mistake in HPAK_AddLump

theAsmodai:


IgnacioFDM:
I believe this fixes this
https://github.com/dreamstalker/rehlds/issues/114

theAsmodai:
It fixes only crash. Sprays are not tested.

Issue #130, opened 2016-01-13
Prevent server/client from sending/requesting same file twice


Jozodezokokotar:
Hi guys,

just want to ask you if you can implement a solution to avoid this issue?

I've seen this as a request and I encountered the issue myself.
https://github.com/ValveSoftware/halflife/issues/1345

Thx!

WPMGPRoSToTeMa:
In .res file there are should be only resources, that can't be precached by map entities.

Jozodezokokotar:
Well, that's not just about .res files. The problem in original hlds is that same resource might be cached by server many times, thus sending/requesting it to client x times.
I might be wrong, but it seems that server does not check (for example if it reads resources from .res files) whether those files are already cached or not.

theAsmodai:
Done.

Jozodezokokotar:
Wunderbar!
Many thanx!

PR #145, opened 2016-02-01
Added preventing twice resource precache in SV_CreateGenericResources… (#130)

WPMGPRoSToTeMa:
… (#130)


WPMGPRoSToTeMa:
Or better move this checks in PF_precache_generic_I function?

theAsmodai:
> Or better move this checks in PF_precache_generic_I function?
Yes, it will be more universal.

WPMGPRoSToTeMa:
@theAsmodai there is mistake in condition, should be:
```C++
	if ((isSoundPrefixed && SV_LookupSoundIndex(&s[soundPrefixLength])) ||
	    SV_LookupModelIndex(s))
```

Issue #154, opened 2016-02-02
Add possibility to write multiline kick reasons


WPMGPRoSToTeMa:
I want something like this:
`kick #1 \nUpdate anti-cheat, you can download it here:\nhttp://github.com/`
will show:
```
You have been disconnected from the server.
Reason: Kicked :
Update anti-cheat, you can download it here:
http://github.com/
```

PR #155, opened 2016-02-02
Added replacing escape sequences in kick reason (#154)

WPMGPRoSToTeMa:
#154


WPMGPRoSToTeMa:
Or better add this feature to all commands?

theAsmodai:
For what exactly commands?

WPMGPRoSToTeMa:
@theAsmodai I think about replacing in Cmd_ExecuteString.

theAsmodai:
No, I think it's not a good idea for incoming commands.

PR #156, opened 2016-02-03
Moved sv_force_ent_intersection in REHLDS_FIXES

WPMGPRoSToTeMa:


WPMGPRoSToTeMa:
@theAsmodai but this cvar doens't exist in original HLDS.

theAsmodai:
Okay.

PR #159, opened 2016-02-05
Added CIDR support for addip, removeip, listip

WPMGPRoSToTeMa:
Fixed invalid mask in StringToFilter for A.B.0.D, A.0.B.D etc.
Added ban time validation in addip
Added optional argument for removeip: removeAll to remove all filters that input filter includes
Added optional argument for listip: ipaddress to show filters which it includes


WPMGPRoSToTeMa:
What you think about additional fields in ipfilter_t such as CIDR?

theAsmodai:
Ok, with REHLDS_FIXES.

PR #178, opened 2016-02-18
Fix GetBonePosition angles

jonatan1024:
GetBonePosition doesn't invert entity's pitch when dealing with its angles.


IgnacioFDM:
But for most mods you can benefit from enabling REHLDS_FIXES, with this change, you have to outright disable all fixes for certain mods.

theAsmodai:
I think we need add a cvar sv_rehlds_fix_pitchangle <1/0> which will allow a complete bug fix in VectorAngles. Then it can be turned on for fixed mods.

LevShisterov:
Possibly better to use liblist.gam parameter in such case.

souvikdas95:
Can anybody explain what this fix is about? Will it produce any significant change? I am currently using mod cstrike along with CS_ReGameDLL.

theAsmodai:
It's only for the coding convenience.

PR #181, opened 2016-02-26
Added auto precaching detail textures if present

WPMGPRoSToTeMa:
Moved auto WAD precaching to TEX_InitFromWad


LevShisterov:
Just wondering, why do you do this on the server-side?

PR #225, opened 2016-07-22
Some fixes for sv_upld.cpp

a1batross:
Hide unneeded checks under REHLDS_FIXES macro.
Use SV_ClearResourceLists() function instead of two SV_ClearResourceList().
Use for() on lists, instead of while(). More readable.

### DIFF
--- a/rehlds/engine/sv_upld.cpp
+++ b/rehlds/engine/sv_upld.cpp
@@ -33,7 +33,7 @@
 // Checks MD5 of the resource against local cache and returns TRUE if resource was found or if downloads are disabled. Otherwise, if resource was requested from the player, it returns FALSE.
 qboolean SV_CheckFile(sizebuf_t *msg, char *filename)
 {
-	resource_t p = { 0 };
+	resource_t p = { };
 	
 #ifdef REHLDS_FIXES
 
@@ -401,7 +401,12 @@ void SV_ParseResourceList(client_t *pSenderClient)
 
 	total = MSG_ReadShort();
 
+#ifdef REHLDS_FIXES
 	SV_ClearResourceLists( host_client );
+#else // REHLDS_FIXES
+	SV_ClearResourceList( &host_client->resourcesneeded );
+	SV_ClearResourceList( &host_client->resourcesonhand );
+#endif // REHLDS_FIXES
 
 #ifdef REHLDS_FIXES
 	if (total > 1) // client uses only one custom resource (spray decal)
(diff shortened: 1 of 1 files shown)


WPMGPRoSToTeMa:
> Use SV_ClearResourceLists() function instead of two SV_ClearResourceList().

Original engine use two SV_ClearResourceList, so put it in REHLDS_FIXES

a1batross:
It's isn't inlined by agressive optimizations?
23.07.2016 1:11 пользователь "Artem Golubikhin" <notifications@github.com>
написал:

rehlds/engine/sv_upld.cpp
```diff
@@ -33,10 +33,8 @@
 // Checks MD5 of the resource against local cache and returns TRUE if resource was found or if downloads are disabled. Otherwise, if resource was requested from the player, it returns FALSE.
 qboolean SV_CheckFile(sizebuf_t *msg, char *filename)
 {
-	resource_t p;
-
-	Q_memset(&p, 0, sizeof(p));
-
+	resource_t p = { 0 };
```
theAsmodai: This line initializes only the first element of structure.

rehlds/engine/sv_upld.cpp
```diff
@@ -259,19 +254,15 @@ void SV_ClearResourceList(resource_t *pList)
...
 	resource_t *p, *n;
 
-	p = pList->pNext;
-	while (p != NULL)
+	for( p = pList->pNext; p && p != pList; p = n )
 	{
-		if (p == pList)
-			break;
-
 		n = p->pNext;
```
theAsmodai: You can move it inside to the loop declaration.

rehlds/engine/sv_upld.cpp
a1batross: Nope. It initializes the first to 0, and the next is 0 too. 

If I placed = { 5 } here, it will initialize first element to 5 and next to 0. 

It's comes from C.

Anyway, there is a C++ so it's possible to initialize struct with empty brace list.

theAsmodai: Ok.

WPMGPRoSToTeMa:
> It's isn't inlined by agressive optimizations?

No because there are 2 calls in 4554 swds (old msvc don't optimize something like this), in 6153 swds too. Also there is check for nullptr (in `SV_ClearResourceLists`).

rehlds/engine/sv_upld.cpp
WPMGPRoSToTeMa: I think empty braces will be more beauty.

rehlds/engine/sv_upld.cpp
souvikdas95: It's so great to see some open discussion on C++ concepts :V

rehlds/engine/sv_upld.cpp
IgnacioFDM: MSVC is dumb enough to optimize = { 0 } and = { } differently 

{ } is faster.

rehlds/engine/sv_upld.cpp
a1batross: It will be less readable.

a1batross:
Done.

rehlds/engine/sv_upld.cpp
theAsmodai: Why?

PR #229, opened 2016-07-29
Added sv_rehlds_userinfo_transmitted_fields cvar

WPMGPRoSToTeMa:


rehlds/engine/info.cpp
```diff
@@ -618,3 +618,37 @@ qboolean Info_IsValid(const char *s)
...
+		const char *value = Info_ValueForKey(srcInfo, key);
+
+		if (value[0] == '\0')
+			continue;
+
+		if (!Q_strcmp(key, "*hltv")
+		 || !Q_strcmp(key, "bottomcolor")
+		 || !Q_strcmp(key, "topcolor"))
+		{
+			if (!Q_atoi(value))
```
theAsmodai: ???

WPMGPRoSToTeMa: This is integer fields, 0 is default value for them. So if they have value equal 0, we can to not send this keys.

theAsmodai: It's looking strange. Why only this three fields? And you can also strip all non-digit chars.

WPMGPRoSToTeMa: > Why only this three fields?

Because only these fields are parsed by svc_updateuserinfo handler.
> And you can also strip all non-digit chars.

Yeah, that is a good idea.

PR #230, opened 2016-07-30
Cvars for command flood punishments

spark512:
Could we get something like this in the next build?


rehlds/rehlds/rehlds_security.cpp
```diff
@@ -56,8 +61,13 @@ void CMoveCommandRateLimiter::CheckBurstRate(unsigned int clientId) {
 		dt = 0.2; //small intervals may give too high rates
 	}
 	if ((m_CurrentMoveCmds[clientId] / dt) > sv_rehlds_movecmdrate_max_burst.value) {
-		Cbuf_AddText(va("addip %i %s\n", 5, NET_BaseAdrToString(cl->netchan.remote_address)));
-		SV_DropClient(cl, false, "Banned for move commands flooding (burst)");
+		if(sv_rehlds_movecmdrate_burst_punish.value < 0) {
+			SV_DropClient(cl, false, "Kicked for move commands flooding (burst)");
+		}
+		else {
+			Cbuf_AddText(va("addip %i %s\n", sv_rehlds_movecmdrate_burst_punish.value, NET_BaseAdrToString(cl->netchan.remote_address)));
```
theAsmodai: You must use %.1f here instead of %i.

spark512:
You're right, updated.

theAsmodai:
Thanks. And please add these cvars description to readme.md

IgnacioFDM:
It would be nice to have a value (maybe -2) to "do nothing, just log"

Issue #239, opened 2016-09-04
ReHLDS and AMXX Files Check


spark512:
Hello,

I'm having a problem with file consistency checks. I'm using the AMXX plugin Files Check (https://forums.alliedmods.net/showthread.php?p=595209) to enforce model consistency, the plugin works fine with the pure version of ReHLDS, but it's misbehaving on the bugfixed version. It's giving an inconsistency error on all listed models, even tho they are the correct ones.

Is there a plugin that forces file consistency like the AMXX Files Check which is compatible with the bugfixed version of ReHLDS?

Mistrick:
For filecheck use [this](http://www.dedicated-server.ru/vbb/showthread.php?t=26550) but only russian description.

spark512:
> For filecheck use this but only russian description.

I am already using ReChecker, but for checking hack files. The AMXX Files Check checks the model's size (mins, maxs) to make sure the client model is an exact match to the server model. ReChecker doesnt do that, ReChecker checks the file's name and hash.

spark512:
The issue is fixed, tested and working on windows. Thank you @WPMGPRoSToTeMa and @theAsmodai .

PR #241, opened 2016-09-08
Fixed optimization-dependent model bounds computation in R_GetStudioB…

WPMGPRoSToTeMa:
…ounds (actually in R_StudioCalcBonePosition) (related to #239)

Tested only under windows, anybody please test it under linux.

### DIFF
--- a/rehlds/engine/r_studio.cpp
+++ b/rehlds/engine/r_studio.cpp
@@ -434,11 +434,19 @@ void R_StudioCalcBonePosition(int frame, float s, mstudiobone_t *pbone, mstudioa
 				// and there's more data in the span
 				if (panimvalue->num.valid > k + 1)
 				{
+#ifdef REHLDS_FIXES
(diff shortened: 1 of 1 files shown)


LevShisterov:
What the purpose of using volatile variables?

WPMGPRoSToTeMa:
> What the purpose of using volatile variables?

That was the first idea how to fix this (suggested by @theAsmodai)

Now I understand that the problem was in `short` to `float` casts.

IgnacioFDM:
I don't understand. The original code should always return the same value regardless of any optimization as far as I understand. How are you testing this?

theAsmodai:
The problem is caused by different floating point calculation precision with different optimization settings.

WPMGPRoSToTeMa:
@IgnacioFDM read #239 to reproduce. In original code `mstudioanimvalue_t::value` can be casted to `float` or `long double`. With `SSE2` in `MSVC` it will cast it to `float`, therefore all calculations will produce `float`. Client code of this function has `long double` cast, `long double` is more accurate than `float`. Thats why client bounds calculations not equal to server bounds calculations.

WPMGPRoSToTeMa:
As I can see via dissasembler the linux engine is fine too.

PR #249, opened 2016-10-20
Cvar Utilities Compatiblity with ReHLDS

NewZMLife:
if i start server while running Cvar Utilities by arkshine, server crashes.
can someone rewrite this plugin or atleast compatible it with ReHLDS ?

https://forums.alliedmods.net/showthread.php?t=154642


Arkshine:
I suggest you don't this module, it contains probably a lot of shit code.  This would need to be rewritten.
But well, most of features are available in AMXX 1.8.3-dev. 
Maybe you should take a look at ReAMXX as well.

PR #295, opened 2016-12-28
Fixed condition in Host_Quit_Restart_f

Chuvi-w:
(This doesn't fix crash when using _restart command in hlds.exe, but at least it fix crash in Host_Quit_Restart_f)

### DIFF
--- a/rehlds/engine/host_cmd.cpp
+++ b/rehlds/engine/host_cmd.cpp
@@ -41,7 +41,7 @@ int gHostSpawnCount;
 qboolean g_bMajorMapChange;
 
 int g_iQuitCommandIssued;
-char *g_pPostRestartCmdLineArgs;
+char *g_pPostRestartCmdLineArgs=NULL;
 
 /*
 * Globals initialization
@@ -498,7 +498,7 @@ void Host_Quit_Restart_f(void)
 	giActive = DLL_RESTART;
 	giStateInfo = 4;
 
-	if (g_psv.active || (g_pcls.state == ca_active && g_pcls.trueaddress[0] && g_pPostRestartCmdLineArgs))
+	if ((g_psv.active || (g_pcls.state == ca_active && g_pcls.trueaddress[0])) && g_pPostRestartCmdLineArgs)
 	{
 		Q_strcat(g_pPostRestartCmdLineArgs, " +connect ");
 		Q_strcat(g_pPostRestartCmdLineArgs, g_pcls.servername);


LevShisterov:
Here how it should look:
```
void Host_Quit_Restart_f(void)
{
	giActive = DLL_RESTART;
	giStateInfo = 4;

	if (g_psv.active)
	{
		if (g_psvs.maxclients == 1 && g_pcls.state == ca_active && g_pPostRestartCmdLineArgs)
		{
			Cbuf_AddText("save quick\n");
			Cbuf_Execute();
			Q_strcat(g_pPostRestartCmdLineArgs, " +load quick");
		}
	}
	else if (g_pcls.state == ca_active && g_pcls.trueaddress[0] && g_pPostRestartCmdLineArgs)
	{
		Q_strcat(g_pPostRestartCmdLineArgs, " +connect ");
		Q_strcat(g_pPostRestartCmdLineArgs, g_pcls.servername);
	}
}
```

rehlds/engine/host_cmd.cpp
```diff
@@ -41,7 +41,7 @@ int gHostSpawnCount;
 qboolean g_bMajorMapChange;
 
 int g_iQuitCommandIssued;
-char *g_pPostRestartCmdLineArgs;
+char *g_pPostRestartCmdLineArgs=NULL;
```
LevShisterov: Doesn't looks like you have to init it. It will have NULL by default. There is a lot of such globals around. Just an IMHO.

Chuvi-w: It's C/C++.
We have not "default" values there, like in pawn.

(В принципе, там да, скорее всего будет 0. Но насколько я помню, в стандарте ничего про это не сказано. Операционка просто выдаёт нам кусок памяти - а что там до этого было - её волновать не должно. Опять-таки, могу ошибаться).

Sorry for russian. xD)

LevShisterov: http://c0x.coding-guidelines.com/6.7.8.html

>  1653 If an object that has static storage duration is not initialized explicitly, then:
>  1654 —  if it has pointer type, it is initialized to a null pointer; 

(If OS will not erase memory pages before giving them to other apps, it will be security hole of the ocean size)

WPMGPRoSToTeMa: http://en.cppreference.com/w/cpp/language/zero_initialization

LevShisterov:
Fixed via #298.

PR #303, opened 2017-01-05
Default *pfilename in COM_Log to status.log

In-line:


rehlds/engine/common.cpp
```diff
@@ -2357,8 +2357,12 @@ void COM_Log(char *pszFile, char *fmt, ...)
 
 	if (!pszFile)
 	{
+#ifdef REHLDS_FIXES
+		pfilename = "status.log";
```
theAsmodai: why not hllog.txt?

In-line: I searched across ReHLDS files, COM_Log has only one usage  
./engine/host_cmd.cpp:535:	COM_Log(szfile, "%s", string);
```
	if (log)
	{
		Q_snprintf(szfile, sizeof(szfile), "%s", "status.log");
		COM_Log(szfile, "%s", string);
	}
```
Do you think that hllog.txt is more suitable ?

theAsmodai: Yes, because `status.log` usable only for one function. We need more universal name.

PR #304, opened 2017-01-06
SetCStrikeFlags() once on server start

In-line:


theAsmodai:
Maybe Host_InitializeGameDLL?

LevShisterov:
Isn't it too late? I think these should be initialized before any code executed in game dll.

PR #312, opened 2017-01-20
Add some C++11 shit to ReHLDS. xD)

Chuvi-w:
Reworked cmd_functions. Test requered.
Added REHLDS_API function to get registered commands.

### DIFF
--- a/rehlds/public/rehlds/rehlds_api.h
+++ b/rehlds/public/rehlds/rehlds_api.h
@@ -33,6 +33,7 @@
 #include "FlightRecorder.h"
 #include "interface.h"
 #include "model.h"
+#include <vector>
 
 #define REHLDS_API_VERSION_MAJOR 3
 #define REHLDS_API_VERSION_MINOR 2
--- a/rehlds/public/rehlds/rehlds_api.h
+++ b/rehlds/public/rehlds/rehlds_api.h
@@ -35,7 +35,7 @@
 #include "model.h"
 
 #define REHLDS_API_VERSION_MAJOR 3
-#define REHLDS_API_VERSION_MINOR 1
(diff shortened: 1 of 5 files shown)


LevShisterov:
This is questionable.
Personally I prefer to maintain existing coding style and base, unless there is no other way.
Why you need commands list, btw?

Chuvi-w:
Something goes wrong.

souvikdas95:
I will let you know the results of text by 22nd Jan.

LevShisterov:
> Command list is requered for planned software.
This is assumed by default. I am interested in the planned usage of this data.

Chuvi-w:
And... hm, I also need a cvars list.
This is requered for tab-completion support.)

LevShisterov:
Then just use Cmd_CompleteCommand or Cmd_GetFirstCmd, no need to use vectors, strings and mem allocation stuff.

rehlds/public/rehlds/rehlds_api.h
```diff
@@ -283,6 +284,7 @@ struct RehldsFuncs_t {
 	void(*SV_UpdateUserInfo)(IGameClient *pGameClient);
 	bool(*StripUnprintableAndSpace)(char *pch);
 	void(*Cmd_RemoveCmd)(char *cmd_name);
+	size_t (*Cmd_GetCmds)(std::vector <std::string> &list);
```
WPMGPRoSToTeMa: You can't use std containers like `vector` and `string` as arguments/returnvalue in external functions, because there are no ABI guarantees, so this is UB. You should create special vector and string classes or use C-style arrays and strings.

theAsmodai:
I think it's not needed.

PR #315, opened 2017-01-21
Remove useless dev mode from Host_WriteConfiguration

In-line:


LeninChan:
if (!developer.value
 +#ifndef REHLDS_FIXES
 +		|| !FS_FileExists("../goldsrc/dev_build_all.bat"
 +#endif
 +		)

maybe?

In-line:
I think it is code that original hlds developers forget to remove after some testing.

theAsmodai:
It's clientside code.

In-line:
@theAsmodai , so why don't fix it ? I think , I don't understand something. Can you explain?

theAsmodai:
We can fix, but it makes no sense. It willn't be used.

PR #318, opened 2017-01-22
Micro optimization and report recursive shutdown by Sys_Error

In-line:


rehlds/engine/host.cpp
```diff
@@ -1008,20 +1008,12 @@ int Host_Frame(float time, int iState, int *stateInfo)
...
-		Cvar_SetValue("violence_agibs", 0.0);
-	}
-	else
-	{
-		Cvar_SetValue("violence_hblood", 1.0);
-		Cvar_SetValue("violence_hgibs", 1.0);
-		Cvar_SetValue("violence_ablood", 1.0);
-		Cvar_SetValue("violence_agibs", 1.0);
-	}
+	float fValue = float(!bLowViolenceBuild);
```
LevShisterov: You should do it so:
`float fValue = bLowViolenceBuild ? 1.0f : 0.0f;`

LevShisterov:
Why do you change printf into Sys_Error?

rehlds/engine/host.cpp
In-line: @LevShisterov I think there is no much difference. What's the point ?

In-line:
@LevShisterov I think this is rare error case. It's not normal when Host_Shutdown called recursively. It is signal that something gone wrong.

rehlds/engine/host.cpp
LevShisterov: To not depend on the value in the variable. Often bXxx variables are of BOOL (int) type, so I didn't checked that this is bool actually. But still, this (:?) form is more clear and right in the result.

LevShisterov:
Both checks (in Host_Error and Host_Shutdown) are used to prevent infinite loops. But here, in Host_Shutdown, it did to do just ignore. You are changing to stop with alerting. So it could change behavior. And I am interested in which cases it would happen and why you was motivated to change it.

In-line:
I reviewed some of engine/ folder sources and tried to find places that can be improved. For me that part was suspicious, I searched across rehlds sources and found no similar approach (with printf).
I do a little more deeper investigation and found, that
1. Host_Error use Sys_Error instead of printf
2. This message is rare case (I was unable to find anything about it in net)
3. I was unable to find in code situation, where it can be produced.
I came to conclusion, that this is some kind of paranoia check or logging added by hlds developers in product early state (to fix it later).

rehlds/engine/host.cpp
In-line: Also this is opposite condition,
`float fValue = bLowViolenceBuild ? 0.0f : 1.0f;`. I will commit it (have no access to PC now), if you want so.

theAsmodai:
Try to run profiler before optimizing anything. The most of optimizations are practically useless.

In-line:
@theAsmodai, this is not just optimization, it also remove copy-paste thing :)

rehlds/engine/host.cpp
LevShisterov: Yes, would be nice.

LevShisterov:
Btw Host_Error check is not against recursive call, but against multiple calls (flag is not cleared at the end).
I see only one sense to change this into Sys_Error - if you wish to try to fix multiple calls. But this function call originate from the launcher, so you actually can't fix it easily. So, honestly, I don't think this makes sense to change that. But, I agree that printf doesn't has a lot meaning here too.

In-line:
@LevShisterov So, program (aka hlds) will be unloaded in any way(with Sys_Error, or with printf). Do you think that launcher has bug with multiple Host_Shutdown calls? I am unable to reproduce it in naive way.

LevShisterov:
Sys_Error leads to crash, printf with return - no. So, in the current variant program can continue. For example _restart leads to Host_Shutdown call and the program should continue after that even if there will be multiple calls to Shutdown. Other code in the Shutdown path looks safe for multiple calls, for example (didn't checked thoroughly though).

In-line:
@LevShisterov Okey.

PR #319, opened 2017-01-22
Fixed auto wad precaching

WPMGPRoSToTeMa:


rehlds/engine/textures.cpp
```diff
@@ -112,7 +112,7 @@ qboolean TEX_InitFromWad(char *path)
 			continue;
 
 #ifdef REHLDS_FIXES
-		if (g_psv.active
+		if (g_psv.edicts != nullptr
```
LevShisterov: Even though it could be logically right, it will be harder to find references in the project and understand what happen when. Isn't there other way? May be `|| g_psv.state == ss_loading` could fit?

WPMGPRoSToTeMa: Oh, yeah, checking state is better. We can just check for `!= ss_dead`, what do you think about it?

LevShisterov: I don't see any assign with ss_dead, so state never reset to it. And I don't think that precaching can be done always after some point.
All variants of restarting map shows that g_psv.state == ss_loading is ok at this point.

WPMGPRoSToTeMa: @LevShisterov it is impicitly resetted when `sv` is resetted via `memset`.

souvikdas95:
What is the effect of this on Server?

spark512:
This update is important, players cannot download wad files from the server...

LevShisterov:
May I ask how it did worked in the past, if it ever was? Was it broken with some change recently?

spark512:
It wasnt working at all, maps with wad files weren't able to be played, we thought it was improperly set fastdl, but it turned out the problem was in rehlds. It' broken on all 0.2 snapshots on nexus and all 3.0 dev releases too.

LevShisterov:
This change is right because PF_precache_generic_I use the same condition:
`if (g_psv.state == ss_loading)`

PR #321, opened 2017-01-23
Allowed spectator and global unreliable buffer to overflow

WPMGPRoSToTeMa:
Extended spectator buffer

### DIFF
--- a/rehlds/engine/server.h
+++ b/rehlds/engine/server.h
@@ -161,6 +161,7 @@ struct rehlds_server_t {
 	// Extended net buffers
 	uint8_t reliableDatagramBuffer[NET_MAX_PAYLOAD];
 	uint8_t signonData[NET_MAX_PAYLOAD];
+	uint8_t spectatorBuffer[MAX_DATAGRAM];
 
 	// Extended resource list
 	resource_t resources[RESOURCE_MAX_COUNT];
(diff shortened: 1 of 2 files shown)


souvikdas95:
What are the causes and effects of this?

WPMGPRoSToTeMa:
It isn't investigated fully, but gamedll sometimes can send a lot of message for `hltv` from player's `TakeDamage`. (when `TakeDamage` is called a lot of times in the same frame) It is good to extend `hltv` buffer to max datagram size to prevent overflows. And 'hltv' buffer was allowed to overflow because engine already has code that tells: `hltv` buffer can overflow, see https://github.com/dreamstalker/rehlds/blob/fb3ec088e1238237fd23b9005309dbb297c8a5c5/rehlds/engine/sv_main.cpp#L4789-L4793
But this not actually solves the problem, there is should be an investigation how to prevent this flood in gamedll (can these messages be combined in one for the same frame?).

souvikdas95:
Is it only related to HLTV and not players?

WPMGPRoSToTeMa:
@souvikdas95 `spectator` buffer is only for HLTV, but it crashes server when overflow.

souvikdas95:
seems to have fixed by crashes though not related to hlds since i have particularly blocked hltv clients from entering the server.

WPMGPRoSToTeMa:
It will crash even without HLTV.

PR #325, opened 2017-01-27
Workaround for texture precaching on case-sensitive filesystem

WPMGPRoSToTeMa:
That fixes a problem when texture file for model ends with "t.mdl", not "T.mdl"

### DIFF
--- a/rehlds/engine/sv_main.cpp
+++ b/rehlds/engine/sv_main.cpp
@@ -5663,7 +5663,7 @@ void PrecacheModelTexture(const char *s, studiohdr_t *pStudioHeader)
 		return;
 
 	size_t modelNameLength = Q_strlen(s);
-	if (modelNameLength >= MAX_QPATH - 1)
+	if (modelNameLength > MAX_QPATH - 2)
 		return;
 
 	char textureModelName[MAX_QPATH];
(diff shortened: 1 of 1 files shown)


rehlds/engine/sv_main.cpp
```diff
@@ -5673,6 +5673,11 @@ void PrecacheModelTexture(const char *s, studiohdr_t *pStudioHeader)
 	char *modelExtension = &textureModelName[modelNameLength - modelExtensionLength];
 	Q_strcpy(modelExtension, "T.mdl");
 
+#ifndef _WIN32
+	if (!FS_FileExists(textureModelName))
+		modelExtension[0] = 't';
```
LevShisterov: Can't target code a way above, but please fix this:
` 	if (modelNameLength >= MAX_QPATH - 1)
return;`
Should be -2, because of `Q_strcpy(modelExtension, "T.mdl");`

WPMGPRoSToTeMa: It's ok, with this condition `MAX_QPATH - 2` is max allowed length of string, so after processing length of string will be `MAX_QPATH - 1` which is ok.

LevShisterov: Ah, yes. Can you change to `> ... -2` ? Thanks.

WPMGPRoSToTeMa: Changed now.

PR #327, opened 2017-01-27
Fixed: parameter constraints in meta forwards & cmd functions

souvikdas95:
This is only with respect to pull request (https://github.com/dreamstalker/metamod-p/pull/1).
This will sync the constraints associated with parameters, properly and in sync with metamod-p.


souvikdas95:
Looks like I'll be damned. Got to work harder...

LevShisterov:
@souvikdas95, can you fix it?

souvikdas95:
@LevShisterov - check now

LevShisterov:
Please do only `const` modifications in this PR and move other changes in separate one.

rehlds/engine/delta.cpp
```diff
@@ -39,7 +39,7 @@ typedef struct delta_link_s
 typedef struct delta_definition_s
 {
 	char *fieldName;
-	int fieldOffset;
```
souvikdas95: @LevShisterov, Is this change ok?

LevShisterov: No, better move it too.

theAsmodai:
Maybe we need to fix 3-rd party mods instead of engine?

souvikdas95:
@theAsmodai, Sarcasm?

LevShisterov:
Its all about warnings in compilations. Shouldn't change actual behavior, but can make code less error prone.

rehlds/engine/pr_cmds.cpp
```diff
@@ -1710,9 +1710,12 @@ int EXT_FUNC PF_DecalIndex(const char *name)
...
 }
 
-void EXT_FUNC PF_lightstyle_I(int style, char *val)
+void EXT_FUNC PF_lightstyle_I(int style, const char *val)
 {
-	g_psv.lightstyles[style] = val;
+	int len = Q_strlen(val) + 1;
+	g_psv.lightstyles[style] = (char*)Hunk_Alloc(len);
+	Q_strncpy(g_psv.lightstyles[style], val, len);
+	g_psv.lightstyles[style][len] = 0;
```
LevShisterov: This one should be in REHLDS_FIXES define, so, move it out of this PR too. I mean Hunk_Alloc usage and surroundings.

souvikdas95: that would require refactoring other patches related to lightstyles in this PR as well as the one in Metamod. will that be fine? as you see, it also uses const but g_psv.lightstyles[] is not of type const char*. So, results in mistype.

LevShisterov: I see no problem in changing server_s.lightstyles into const what it actually should be, because it can't be freely modified.

souvikdas95: https://github.com/dreamstalker/rehlds/blob/master/rehlds/engine/host_cmd.cpp#L1842-L1843

LevShisterov: No problem with that too:

> 			char *val = (char *)Hunk_Alloc(Q_strlen(light.style) + 1);
>			Q_strncpy(val, light.style, 3);
>			val[3] = 0;
>			g_psv.lightstyles[light.index] = val;

souvikdas95: Shall I include that in this PR or in a new PR?

souvikdas95: @LevShisterov , done. Check now.

LevShisterov:
Thanks!

PR #331, opened 2017-01-30
Small fixes and formatting.

LevShisterov:


souvikdas95:
Can you also address the issue of failing to start hlds.exe from outside working directory? For example, on windows, your hlds location is "C:\testhlds\hlds.exe" and you want to start it directly from command prompt (which uses C:\ as default working directory for non administrator and "C:\windows\System32" for administrator. As a result, HLDS crashes because it fails to find required .wad files to start up the server.

In-line:
@souvikdas95 I think you can do something like this https://stackoverflow.com/questions/9597001/how-to-set-the-working-directory-of-a-command-in-a-windows-batch-file

https://stackoverflow.com/questions/3161989/how-do-i-set-the-current-working-directory-drive-in-a-dos-batch-file

souvikdas95:
@In-line I know we can use batch files to do these jobs. But in some windows debugging tools, setting working directory is not possible. If we use batch files, it assumes that cmd.exe used for that is the file in question to debug and not hlds.exe.

souvikdas95:
Good job. @LevShisterov now a Collaborator!

PR #339, opened 2017-02-06
GCC support

LevShisterov:


rehlds/engine/cmd.cpp
```diff
@@ -1010,7 +1010,7 @@ qboolean Cmd_ForwardToServerInternal(sizebuf_t *pBuf)
 	char tempData[4096];
 	sizebuf_t tempBuf;
 
-	tempBuf.buffername = __FUNCTION__ "::tempBuf";
+	tempBuf.buffername = "Cmd_ForwardToServerInternal::tempBuf";
```
WPMGPRoSToTeMa: `va("%s::tempBuf", __FUNCTION__)`?
And maybe add something like `nameof` for tempBuf variable name in string?
[Something like this](https://gist.github.com/Terik23/aeba961ecc870d6cd01404b4ab989d79)
So:
`tempBuf.buffername = va("%s::%s", __FUNCTION__, nameof_variable(tempBuf));`

theAsmodai: constexpr isn't supported by current visual studio

WPMGPRoSToTeMa:
I think we should use `__func__` instead of `__FUNCTION__` because latter is not standardized.

WPMGPRoSToTeMa:
It seems we should define `__func__` to `__FUNCTION__` for VS2013.

WPMGPRoSToTeMa:
Good job!

rehlds/engine/pr_cmds.cpp
```diff
@@ -324,13 +324,13 @@ void EXT_FUNC PF_ambientsound_I(edict_t *entity, float *pos, const char *samp, f
 void EXT_FUNC PF_sound_I(edict_t *entity, int channel, const char *sample, float volume, float attenuation, int fFlags, int pitch)
 {
 	if (volume < 0.0 || volume > 255.0)
-		Sys_Error("EMIT_SOUND: volume = %i", volume);
+		Sys_Error("%s: volume = %i", __func__, volume);
```
WPMGPRoSToTeMa: There is some problem: old code uses `EMIT_SOUND`, but new uses `PF_sound_I`.

LevShisterov: The same as happen to PF_WriteByte_I, etc...

LevShisterov:
> Should all these be in REHLDS_FIXES?

No. It can't break demos.

PR #341, opened 2017-02-10
Fixed volume checking in emit sound

WPMGPRoSToTeMa:
TODO:
- [x] additional info like sound name and entity classname in errors for better debugging.
- [x] rework `LocalCheckBounds`?


rehlds/engine/pr_cmds.cpp
```diff
@@ -323,8 +323,13 @@ void EXT_FUNC PF_ambientsound_I(edict_t *entity, float *pos, const char *samp, f
 
 void EXT_FUNC PF_sound_I(edict_t *entity, int channel, const char *sample, float volume, float attenuation, int fFlags, int pitch)
 {
+#ifdef REHLDS_FIXES
+	if (!(0.0 <= volume && volume <= 1.0))
```
LevShisterov: Please use similar form:
`if (volume < 0.0 || volume > 1.0)`
Thanks.

In-line:
Maybe we don't need crash server with `Sys_Error(...)`. I think it's not such critical error to do that.

spark512:
> Maybe we don't need crash server with Sys_Error(...). I think it's not such critical error to do that.

Yeah, why crash the server due to an improper sound? Just throw an error and ignore/remove the sound, the server can keep going..

WPMGPRoSToTeMa:
Without crash a lot of people will just ignore that error, this is not good, because the reasons of bad arguments can be non-obvious (like we got from regamedll func_breakable), but additional info should help. We can show error(s), clamp arguments in needed range and play sound (if there is no problems with `channel` arg).

In-line:
@WPMGPRoSToTeMa It's good idea too. Why did not you use the lamba with auto parameters? I think it can be more readable.

WPMGPRoSToTeMa:
@In-line you can't stringize expression without macro.

spark512:
Any progress on this one?

In-line:
@WPMGPRoSToTeMa Ping.

WPMGPRoSToTeMa:
Can someone of collaborators remove ignore of this PR in TeamCity?

s1lentq:
Done.

theshadowz:
Is it going to be merged?

rehlds/engine/pr_cmds.cpp
```diff
@@ -312,6 +312,18 @@ void EXT_FUNC PF_ambientsound_I(edict_t *entity, float *pos, const char *samp, f
 
 void EXT_FUNC PF_sound_I(edict_t *entity, int channel, const char *sample, float volume, float attenuation, int fFlags, int pitch)
 {
+#ifdef REHLDS_FIXES
+	auto checkBounds = [&](double varValue, char const *varName, double min, double max) {
```
wopox1337, line 316: Why use `[&]` instead `[=]` for declaration? The context parameters are not changed, only read.

wopox1337 (changes requested):

rehlds/engine/pr_cmds.cpp
wopox1337: Done https://github.com/dreamstalker/rehlds/pull/341/commits/e064ac8e3da6131ca64d45eeb070233ecc88a100

rehlds/engine/pr_cmds.cpp
wopox1337: Done https://github.com/dreamstalker/rehlds/pull/341/commits/cb3cdc59c0ce00183514e84a17cc05586db6e60d

wopox1337 (approved):

WPMGPRoSToTeMa (commented):
@s1lentq which behavior changes are allowed to be not `#ifdef`ed?

wopox1337:
@WPMGPRoSToTeMa Fix is not important. And readability exceeds usefulness #ifdef REHLDS_FIXES

rehlds/engine/sv_main.cpp
```diff
@@ -670,26 +670,21 @@ qboolean SV_BuildSoundMsg(edict_t *entity, int channel, const char *sample, int
 	int sound_num;
 	int field_mask;
 
-	if (volume < 0 || volume > 255)
+	auto clampBounds = [=, func = __func__]<typename T>(T value, char const *name, T min, T max)
```
WPMGPRoSToTeMa, line 673: Pass `value` by ref as you're changing it.

rehlds/engine/pr_cmds.cpp
```diff
@@ -312,16 +312,18 @@ void EXT_FUNC PF_ambientsound_I(edict_t *entity, float *pos, const char *samp, f
 
 void EXT_FUNC PF_sound_I(edict_t *entity, int channel, const char *sample, float volume, float attenuation, int fFlags, int pitch)
 {
-	auto checkBounds = [=](double varValue, char const *varName, double min, double max) {
-		if (varValue < min || varValue > max) {
-			Sys_Error("EMIT_SOUND: %s=%g out of bounds %g-%g\nEntity classname = %s, sound = %s\n",
-				varName, varValue, min, max, entity->v.classname + pr_strings, sample);
+	auto checkBounds = [=, func = __func__]<typename T>(T value, char const *name, T min, T max)
```
WPMGPRoSToTeMa, line 315: Template lambdas are in C++20. As I know this project is targeting C++14.

wopox1337: Yes, you're right. Then I don't know how to do it for C++14.

rehlds/engine/sv_main.cpp
wopox1337: Thanks! Done https://github.com/dreamstalker/rehlds/pull/341/commits/7978f70487f158881a1bd1ce7c4e71e942812689

rehlds/engine/pr_cmds.cpp
WPMGPRoSToTeMa: @wopox1337 you can turn it into a separate static function and pass `__func__`, `entity` and `sample` as parameters.

wopox1337: Thank you! I'll fix it tomorrow.

wopox1337: > Or you can use auto instead, so there will be 3 template parameters.

Then we will lose the type check for the compiler.

wopox1337: Unfortunately, no suitable solution with lambda functions was found. So in this PR we leave only the essence - fixing the error of the original HLDS without refactoring. 🤔

s1lentq (approved):

rehlds/engine/pr_cmds.cpp
```diff
@@ -312,7 +312,11 @@ void EXT_FUNC PF_ambientsound_I(edict_t *entity, float *pos, const char *samp, f
 
 void EXT_FUNC PF_sound_I(edict_t *entity, int channel, const char *sample, float volume, float attenuation, int fFlags, int pitch)
 {
+#ifdef REHLDS_FIXES
+	if (volume < 0.0f || volume > 1.0f)
+#else
 	if (volume < 0.0 || volume > 255.0)
+#endif
 		Sys_Error("%s: volume = %i", __func__, volume);
```
WPMGPRoSToTeMa, line 320: @wopox1337 @s1lentq 
Note that the merged changes doesn't include a fix for this line, it should be:
```C++
Sys_Error("%s: volume = %f", __func__, volume);
```

wopox1337: Already fixed in master branch

rehlds/engine/pr_cmds.cpp
Hun1eR: @wopox1337 Actually you can use auto instead of template parameters. And this is should work on c++14:
```cpp
auto checkBounds = [=, func = __func__](auto value, char const *name, decltype(value) min, decltype(value) max)
```

WPMGPRoSToTeMa: @Hun1eR your suggestion looks great!

PR #352, opened 2017-02-14
Remove comment

In-line:


rehlds/engine/world.cpp
```diff
@@ -181,7 +181,7 @@ struct hull_s *SV_HullForBsp(edict_t *ent, const vec_t *mins, const vec_t *maxs,
 			else
 				hull = &model->hulls[3];
 		}
-// TODO: add cvar for this?
+// DONE: add cvar for this
```
WPMGPRoSToTeMa: I have forgotten what I meant by this.

rehlds/engine/world.cpp
```diff
@@ -181,7 +181,7 @@ struct hull_s *SV_HullForBsp(edict_t *ent, const vec_t *mins, const vec_t *maxs,
 			else
 				hull = &model->hulls[3];
 		}
-// TODO: add cvar for this?
```
WPMGPRoSToTeMa: I have forgotten what I meant by this.

In-line: @WPMGPRoSToTeMa I think you meant `sv_rehlds_hull_centering` which was added later

In-line: @WPMGPRoSToTeMa Hm. It was added by this commit fe9bfbb. Anyway it's confusing

WPMGPRoSToTeMa: I think we can just remove this line.

In-line: @WPMGPRoSToTeMa I removed it.

PR #357, opened 2017-02-15
Suffix parameters in type declaration to avoid messing with macros and local variables.

LevShisterov:


WPMGPRoSToTeMa:
I think the problem is in macros, not in argument names, but we can't remove these macros, so it's ok.

PR #366, opened 2017-02-17
[Codebase] const-like defines -> consts

Sh1ft0x0EF:


WPMGPRoSToTeMa:
Pragma once isn't standardized yet.

Sh1ft0x0EF:
And won't be I presume
But it works well and compilers can process headers a bit faster using it instead of guards

Sh1ft0x0EF:
It's a bit strange that win compiler couldn't recognize constexpr
Some of your code is using features of C++11 but

Sh1ft0x0EF:
You probably would be dead before pragma become standardized

rehlds/engine/hashpak.h
```diff
@@ -32,7 +32,8 @@
 #include "custom.h"
 #include "FileSystem.h"
 
-constexpr auto HASHPAK_EXTENSION = ".hpk";
+constexpr char *HASHPAK_EXTENSION = ".hpk";
```
WPMGPRoSToTeMa: Better change `char*` to `const char*` in needed function.

LevShisterov:
Please split this in two: guard and other changes.

Sh1ft0x0EF:
You wanna keep all the constexprs but restore guards? Or pragma with guards? Like that:

```cpp
#pragma once
#ifndef BLABLA_H
#define BLABLA_H
```

WPMGPRoSToTeMa:
Pragma with guards should be good.

Sh1ft0x0EF:
Alright

LevShisterov:
Use two (or more) pull requests for these changes. It will make review easier and accept faster.

Sh1ft0x0EF:
Win build has failed here (precompiled failed to be compiled because of constexpr usage)

In-line:
What are advantages for `constexpr auto`?

IgnacioFDM:
I'd say only using pragma once (without header guards) is better. It reduces the boilerplate, and the chance of making a mistake when writing the guards resulting in potential errors. Also you will never ever find a C++ compiler capable enough to support the C++11 stuff used here that doesn't support pragma once.

Chuvi-w:
What's the main difference between _#define_ and _constexpr auto_?
Project is still compiliable in VS2013. _constexpr_ will break it.

In-line:
@Chuvi-w I think compiler will generate debug info for `constexpr`, but will not for `#define`

Chuvi-w:
@In-line do you really need debug info for thouse defines?

WPMGPRoSToTeMa:
constexpr is type-safe.

Chuvi-w:
@WPMGPRoSToTeMa Типа безопасный?

Sh1ft0x0EF:
Finally!
Replaced const-like defines with consts for now (the only those which could be safely replaced)
I will replace them fully and with constexpr a bit later but this PR can be applied now

rehlds/engine/net.h
```diff
@@ -91,25 +91,25 @@ enum
...
 
 // Flow control bytes per second limits
-#define MAX_RATE		100000.0f
-#define MIN_RATE		1000.0f
+const float MAX_RATE		= 100000.0f;
+const float MIN_RATE		= 1000.0f;
 
 // Default data rate
-#define DEFAULT_RATE	(9999.0f)
+const float DEFAULT_RATE	= (9999.0f);
```
In-line: There is no reason to use braces here

PR #368, opened 2017-02-17
Improve header guards in engine code

Sh1ft0x0EF:


rehlds/engine/common.h
```diff
@@ -37,7 +41,7 @@
 #ifndef REHLDS_FIXES
 #ifdef _WIN32
 	#define __BUILD_TIME__ "11:17:23"
-	#define __BUILD_DATE__ "Aug  8 2013"
+	#define __BUILD_DATE__ "Aug 8 2013"
```
LevShisterov: Revert this.

rehlds/engine/delta.h
```diff
@@ -25,68 +25,71 @@
...
-#define DT_SIGNED			BIT(31)		// sign modificator
+#define DT_BYTE              BIT(0)        // A byte
+#define DT_SHORT             BIT(1)        // 2 byte field
+#define DT_FLOAT             BIT(2)        // A floating point field
+#define DT_INTEGER           BIT(3)        // 4 byte integer
+#define DT_ANGLE             BIT(4)        // A floating point angle
+#define DT_TIMEWINDOW_8      BIT(5)        // A floating point timestamp relative to server time
+#define DT_TIMEWINDOW_BIG    BIT(6)        // A floating point timestamp relative to server time (with more precision and custom multiplier)
+#define DT_STRING            BIT(7)        // A null terminated string, sent as 8 byte chars
+#define DT_SIGNED            BIT(31)       // sign modificator
```
LevShisterov: We are using tabs, not spaces.

rehlds/engine/delta.h
```diff
@@ -25,68 +25,71 @@
...
 typedef struct delta_s delta_t;
 typedef void(*encoder_t)(delta_t *, const unsigned char *, const unsigned char *);
 
 typedef struct delta_stats_s
 {
-	int sendcount;
-	int receivedcount;
+    int sendcount;
+    int receivedcount;
 } delta_stats_t;
```
LevShisterov: Same

rehlds/engine/delta.h
```diff
@@ -25,68 +25,71 @@
...
+    int fieldType;
+    char fieldName[32];
+    int fieldOffset;
+    short int fieldSize;
+    int significant_bits;
+    float premultiply;
+    float postmultiply;
+    short int flags;
+    delta_stats_t stats;
 } delta_description_t;
```
LevShisterov: Same, etc ...

LevShisterov:
Still there are a lot of other changes, not only the guards.

Sh1ft0x0EF:
So I should to undone 'em for now?

rehlds/engine/delta.h
WPMGPRoSToTeMa: @LevShisterov for alignment we should use spaces, but this is for another commit.

LevShisterov:
Okay, let us discuss code formatting a bit.
For the start:

- tabs for indent, but mix with spaces for multi-line (resharper default). [sample](https://github.com/dreamstalker/rehlds/blob/master/rehlds/engine/sv_main.cpp#L7025), though you can see that if your editor doesn't support setting tab size to 4, you will have it unaligned. But we are coding in VS. In browser you can change tab size with the plugin/styles.

- file ending should have a new line. [not_ok](https://github.com/dreamstalker/rehlds/pull/368/files#diff-21ec313554d34c50f411752a3ad1e878R939), [ok](https://github.com/dreamstalker/rehlds/pull/368/files#diff-10d89fa18370c7efa70124e8cab86fbbR108). At least do not change to this.

- alignment in adjacent defines: [looks_better_before_than_after](https://github.com/dreamstalker/rehlds/pull/368/files#diff-21ec313554d34c50f411752a3ad1e878R916) (keep 4 tab size in mind).

- personally I used tabs for values alignment in defines: [same_sample](https://github.com/dreamstalker/rehlds/pull/368/files#diff-21ec313554d34c50f411752a3ad1e878R916), but we can discuss this.

And for this PR: leave only guard changes here so we can easily accept it (looks like we are ok with using double guards and drop support for GCC <3.4 where `#pragma once` isn't supported).
Btw, why not to drop macro-guards and leave only `#pragma once`?

Sh1ft0x0EF:
You should create a wiki page describing your coding style. Would be better instead of describe them to each new person

`In browser you can change tab size with the plugin/styles.`

You can use .editorconfig file to set tabsize to 4 for browser view

Despite that you are using hdr guards not all of your files are using them (there is only pragma in there)
And despite that you want file to have a newline not all your files have it
And yet most important thing is that you don't have any coding style guide for your project (at least around the repo)

About alignment I think that I need to align them in a row with spaces (or they will be looking ugly at tabsize 8)

I think tabs are good and I'm using them too but then you need to create a .editorconfig file in repo that should look somehow like that:
https://github.com/Sh1ft0x0EF/hlsdk-core/blob/master/.editorconfig

Currently pragma once are supported around all modern compilers and there no any problem with it. It is not standardized but it's too long to wait for it (C++17 brings the standard filesystem functionality after 20 years! Maybe they standardize pragma in 2040...)

Sh1ft0x0EF:
Review required

WPMGPRoSToTeMa:
We should use spaces for alignment because it would looks aligned with any tabsize.
> file ending should have a new line.

Why?

Sh1ft0x0EF:
It'll be good to apply clang-format. It still bad at defines values alignment but you can manually disable some parts of code to be formatted. There is also a tool called uncrustify which has more functionality but gives use an opportunity to have code formatted almost exactly like you want

LevShisterov:
@WPMGPRoSToTeMa because of [0](https://github.com/dreamstalker/rehlds/pull/368/files#diff-21ec313554d34c50f411752a3ad1e878R939) and Ctrl+Shift+End puts cursor not in the first column. So it is personal preference.

Ok, I am up for using spaces for alignment, though I think it could be a lot of work.
Also I am up to remove macro-guards and leave only `#pragma once`.

WPMGPRoSToTeMa:
> It is not standardized but it's too long to wait for it

They are trying to standardize modules now. Modules should replace header files, but this can't be done immediately.

Sh1ft0x0EF:
Think that this can be closed now (overridden by #371)

PR #372, opened 2017-02-18
enum-like defines -> enums

Sh1ft0x0EF:


rehlds/engine/sys_engine.h
```diff
@@ -32,9 +32,11 @@
...
 #include "iengine.h"
 
-// sleep time when not focus
-#define NOT_FOCUS_SLEEP	50
-#define MINIMIZED_SLEEP 20
+enum
+{
+	MINIMIZED_SLEEP = 20,
+	NOT_FOCUS_SLEEP = 50 // sleep time when not focus
+};
```
WPMGPRoSToTeMa: These should be just constants.

Sh1ft0x0EF: Ok

PR #374, opened 2017-02-20
Fix IsSafeFileToDownload (thx [WPMG]PRoSToTeM@)

In-line:
* Fix nullptr dereferencing
* Remove redundant check 
* Fix comment

### DIFF
--- a/rehlds/engine/sv_main.cpp
+++ b/rehlds/engine/sv_main.cpp
@@ -7303,7 +7303,7 @@ void SV_FailDownload(const char *filename)
 qboolean IsSafeFileToDownload(const char *filename)
 {
 	char *first;
-	char *last; // TODO: Why do we need first/last ?
+	char *last;
 
 	char lwrfilename[MAX_PATH];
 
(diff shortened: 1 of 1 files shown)


rehlds/engine/sv_main.cpp
```diff
@@ -7327,17 +7327,22 @@ qboolean IsSafeFileToDownload(const char *filename)
...
+	else
+		last = nullptr;
 #else
 	last = Q_strrchr(lwrfilename, '.');
 #endif
 
 	if (lwrfilename[0] == '/'
 		|| Q_strstr(lwrfilename, "\\")
 		|| Q_strstr(lwrfilename, ":")
+#ifndef REHLDS_FIXES // Redundant check
```
theAsmodai: Why?

LevShisterov:
This function should be rewritten from scratch, IMHO. I see no reason to limit ext to 3 chars, require ext, etc..

rehlds/engine/sv_main.cpp
In-line: Later we check dot count to be 1 (I added comment about it)

In-line:
@LevShisterov Are you sure it is safe to do that ?

LevShisterov:
Depends on how good it will be done.

rehlds/engine/sv_main.cpp
theAsmodai: Ok, but such things must be commented

IgnacioFDM:
We would definitely need some very extensive unit-testing coded for this function if any big changes were to happen

rehlds/engine/sv_main.cpp
In-line: Okey. I will later do that

In-line:
@LevShisterov We can add `// TODO: ...` to the code about it.

PR #375, opened 2017-02-20
Add MAX_CVARLIST_FILES define

In-line:


LevShisterov:
No need for REHLDS_FIXES and value checks.

In-line:
@LevShisterov Here will be problem if value is above 100 https://github.com/In-line/rehlds/blob/bd8019da9f01308ef57f957b8614e127bed137f8/rehlds/engine/cvar.cpp#L595

In-line:
@LevShisterov Fixed. My mistake.

LevShisterov:
No problem. Thanks.

PR #380, opened 2017-02-24
Fix #235

In-line:
This sections are almost identical 
https://github.com/dreamstalker/rehlds/blob/master/rehlds/engine/host_cmd.cpp#L570
https://github.com/dreamstalker/rehlds/blob/master/rehlds/engine/host_cmd.cpp#L701
I think author of PR forget to fix second section.

Should I include unused variables in `#ifndef REHLDS_FIXES` ?

### DIFF
--- a/rehlds/engine/host_cmd.cpp
+++ b/rehlds/engine/host_cmd.cpp
@@ -698,12 +698,14 @@ void Host_Status_Formatted_f(void)
 			return;
 		}
 	}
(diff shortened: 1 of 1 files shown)


WPMGPRoSToTeMa:
Client can't send `stat` and `stat log` of course, so there should be no problems. I think we blocked `status log` because of client can flood the server with writing to `status.log`.

PR #386, opened 2017-02-28
Add separate files for MD5

In-line:


Chuvi-w:
This files seems to be missed in vcxproj.

In-line:
https://github.com/In-line/rehlds/blob/8dc893929534426e9393169da1d8872b34af3d60/rehlds/msvc/ReHLDS.vcxproj#L324

Chuvi-w:
.h is a header only, Visual Studio doesn't include existing .cpp automatically.

In-line:
@Chuvi-w My mistake. I can't send PR for that now. Please, do it. I use linux as Desktop OS, so was unable to test.

Chuvi-w:
Maybe some times later. (If nobody noticed this bug, maybe, no one uses Visual Studio 0_o)

LevShisterov:
Fixed.

PR #387, opened 2017-02-28
Use Host_IsSinglePlayerGame whenever possible

In-line:
- [x] Find all singleplayer/multiplayer checks
- [x] Make PR build
- [x] Did we include all in this PR?
- [ ] Is it safe to use Host_IsSinglePlayerGame?

### DIFF
--- a/rehlds/engine/host.cpp
+++ b/rehlds/engine/host.cpp
@@ -1023,18 +1023,9 @@ void CheckGore(void)
 qboolean Host_IsSinglePlayerGame(void)
 {
 	if (g_psv.active)
-#ifdef REHLDS_FIXES
-		return g_psvs.maxclients <= 1;
-#else
 		return g_psvs.maxclients == 1;
-#endif
 	else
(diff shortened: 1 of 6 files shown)


rehlds/engine/host.cpp
```diff
@@ -1019,9 +1023,17 @@ void CheckGore(void)
...
+		return g_psvs.maxclients <= 1;
+#else
 		return g_psvs.maxclients == 1;
+#endif
 	else
+#ifdef REHLDS_FIXES
+		return g_pcl.maxclients <= 1;
+#else
 		return g_pcl.maxclients == 1;
+#endif
```
In-line: @LevShisterov `else` part only?

LevShisterov: All changes in this func.

In-line: @LevShisterov Fixed.

In-line:
@LevShisterov Do you think it's fine in other cases ?

PR #398, opened 2017-03-12
#pragma region test in ICC. Don't merge, plz.

Chuvi-w:
PS: About this stuff: https://msdn.microsoft.com/ru-ru/library/b6xkz944.aspx


Chuvi-w:
Hm. Seems, it works. Thanks.)

PR #403, opened 2017-03-13
Fix tv_usec calculation in NET_Sleep_Timeout

In-line:
This is the original code: https://github.com/dreamstalker/rehlds/blob/7326bee095ca1fcc003842f0f6ed5ec622695853/rehlds/engine/net_ws.cpp#L1094
`(1000 / fps) * 1000` and `1000 * 1000 / fps` are not equivalent.


theAsmodai:
Please recheck it in disassembler. As far as I can remember, there was no mistake.

LevShisterov:
Actually it is `(1000 / fps) * 1000` in original engine.

PR #407, opened 2017-03-19
Allow to modify triggers collection while iterating it (prevent crash).

LevShisterov:


rehlds/engine/pr_cmds.cpp
```diff
@@ -113,7 +113,7 @@ void EXT_FUNC SetMinMaxSize(edict_t *e, const float *min, const float *max, qboo
 	e->v.size[0] = max[0] - min[0];
 	e->v.size[1] = max[1] - min[1];
 	e->v.size[2] = max[2] - min[2];
-	SV_LinkEdict(e, 0);
```
WPMGPRoSToTeMa: Maybe split refactoring changes like this into another commit?

rehlds/engine/pr_cmds.cpp
```diff
@@ -113,7 +113,7 @@ void EXT_FUNC SetMinMaxSize(edict_t *e, const float *min, const float *max, qboo
 	e->v.size[0] = max[0] - min[0];
 	e->v.size[1] = max[1] - min[1];
 	e->v.size[2] = max[2] - min[2];
-	SV_LinkEdict(e, 0);
+	SV_LinkEdict(e, FALSE);
```
WPMGPRoSToTeMa: Maybe split refactoring changes like this into another commit?

theAsmodai: There is no actual need for this.

PR #410, opened 2017-03-19
Launcher refactoring

theAsmodai:


rehlds/common/commandline.cpp
```diff
@@ -321,30 +321,28 @@ const char *CCommandLine::CheckParm(const char *psz, char **ppszValue) const
 		return nullptr;
 
 	char *pret = strstr(m_pszCmdLine, psz);
-	if (!pret || !ppszValue)
-		return nullptr;
+	if (pret && ppszValue) {
```
LevShisterov: Looks like you can invert condition and reduce nesting.

rehlds/dedicated/src/sys_linux.cpp
```diff
@@ -200,14 +246,13 @@ void CSys::Printf(char *fmt, ...)
 }
 
 const int MAX_LINUX_CMDLINE = 2048;
-static char linuxCmdline[ MAX_LINUX_CMDLINE ];
+static char linuxCmdline[MAX_LINUX_CMDLINE];
 
-void BuildCmdLine(int argc, char **argv)
+char* BuildCmdLine(int argc, char **argv)
```
In-line: It doesn't return anything.

rehlds/dedicated/src/sys_window.cpp
```diff
@@ -2,31 +2,41 @@
...
+	char *ConsoleInput() override;
+	void Printf(char *fmt, ...) override;
 };
 
 CSys g_Sys;
 ISys *sys = &g_Sys;
 
+CSys::CSys()
+{
+	// Startup winock
```
In-line: *winsock

PR #433, opened 2017-04-23
Update IsComment

In-line:
1. Refactor
2. Instead of making function call, check condition without it.

### DIFF
--- a/rehlds/engine/tmessage.cpp
+++ b/rehlds/engine/tmessage.cpp
@@ -115,14 +115,11 @@ char* EXT_FUNC memfgets(unsigned char *pMemFile, int fileSize, int *pFilePos, ch
 
 int IsComment(char *pText)
 {
-	int length;
 	if (!pText)
-	{
 		return TRUE;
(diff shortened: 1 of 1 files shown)


WPMGPRoSToTeMa:
Better `pText[0] == '\0'`

rehlds/engine/tmessage.cpp
```diff
@@ -115,17 +115,19 @@ char* EXT_FUNC memfgets(unsigned char *pMemFile, int fileSize, int *pFilePos, ch
...
-	int length;
 	if (!pText)
-	{
 		return TRUE;
-	}
 
-	length = Q_strlen(pText);
-	if (length >= 2 && pText[0] == '/' && pText[1] == '/' || length <= 0)
+#ifdef REHLDS_FIXES
+	if ((pText[0] == '/' && pText[1] == '/') || !pText[0])
```
WPMGPRoSToTeMa: Better `pText[0] == '\0'`

PR #436, opened 2017-04-24
Use standard library in RandomLong/RandomFloat

In-line:
1. `std::random_device` provides seed from entropy, which is better then seeding from `std::time(nullptr)`
2. Code is simplified
3. And of course use of standard library is better.


In-line:
Actually building was successful, TeamCity failed with weird error.

s1lentq:
@In-line oops, my fault, i'll be fix

rehlds/engine/pr_cmds.cpp
```diff
@@ -28,6 +28,10 @@
 
 #include "precompiled.h"
 
+#ifdef REHLDS_FIXES
+#include <random>
```
LevShisterov: This include should go in osconfig.h.

In-line: @LevShisterov One second

In-line: @LevShisterov Done.

WPMGPRoSToTeMa:
Why close?

In-line:
@WPMGPRoSToTeMa This PR requires newer libstdc++ then in HLDS root. So, we should decide to link libstdc++ statically or bypass this problem another way. For now I close this.

PR #439, opened 2017-04-25
Optimize PF_IsMapValid_I

In-line:


rehlds/engine/pr_cmds.cpp
```diff
@@ -1541,11 +1541,15 @@ int EXT_FUNC PF_precache_generic_I(const char *s)
 
 int EXT_FUNC PF_IsMapValid_I(const char *mapname)
 {
+#ifdef REHLDS_FIXES
+	char cBuf[42];
```
theAsmodai: :suspect:

WPMGPRoSToTeMa: Need a comment like:
`[len('maps/')=5] + 32 + [len('.bsp')=4] + [eos=1]`

theAsmodai: It's a very minor optimization. Use of 64 would been easier than to count characters.

WPMGPRoSToTeMa: @theAsmodai yup.

PR #440, opened 2017-04-25
Add project HLTV

s1lentq:
HLTV includes 5 projects that:
- [x] Console
- [x] Core
- [x] Proxy
- [x] DemoPlayer
- [x] Director


In-line:
Welcome ReHLTV, welcome fake players :unamused: !

rehlds/HLTV/Console/src/System.cpp
```diff
@@ -0,0 +1,1188 @@
...
+	float fps = 0.0f;
+	float nextConsoleUpdate = 0.0f;
+
+	BaseSystemModule::RunFrame(time);
+	while ((m_State == MODULE_RUNNING || m_State == MODULE_CONNECTING))
+	{
+		UpdateTime();
+		double timeDiff = m_SystemTime - m_LastTime;
+		if (m_LastTime > 0 && timeDiff <= 0) {
+			Printf("WARNING! System::RunFrame: system time difference <= 0.\n");
```
In-line: Maybe we can use something like c++11 steady_clock ? (For example CLOCK_MONOTONIC in linux)

WPMGPRoSToTeMa: It requires static linking of stl library. (but i don't mind it)

WPMGPRoSToTeMa: @In-line `Sys_FloatTime` already uses `CLOCK_MONOTONIC` in linux.

In-line: @WPMGPRoSToTeMa We need to check win32 implementation too (but it's pain to read it).

WPMGPRoSToTeMa: @In-line it uses QPC which is monotonic (`steady_clock` also uses QPC), but there are some another code after QPC call, need to read it.

SkillartzHD:
http://pvpru.com/board/showthread.php?t=520628

what's wrong with this ? or have necessarily to be a brand own ?

s1lentq:
@SkillartzHD what does it have to do with pr?
it is absolutely different things.

spark512:
You guys are awesome :)

SkillartzHD:
@s1lentq 
do you mean that it is a good idea to free a source code with a player who has netchanel attached and all messages svc_ ? etc..  it is enough just to delete the *hltv in connect

PR #443, opened 2017-04-26
Fix possible null dereferences

In-line:
Found with cppcheck. They aren't 100% null dereferences, but can create dangerous situations in the future.

Inspect this calls:
1. [R_StudioCalcBonePosition](https://github.com/In-line/rehlds/blob/68fcac46e26f364f32f9220c4156a00f7bf7bbab/rehlds/engine/r_studio.cpp#L1097) `(adj == nullptr)`: 

2. [NET_LagPacket](https://github.com/In-line/rehlds/blob/3fd4e45420dfe754bb3bea4fa49d3f0bb62dbce6/rehlds/engine/net_ws.cpp#L1030) (most of params are null)

And [this line of code](https://github.com/In-line/rehlds/blob/558b1e367f10f1587e2fda401dae06bbf809fa09/rehlds/engine/sv_main.cpp#L4308)

### DIFF
--- a/rehlds/engine/net_ws.cpp
+++ b/rehlds/engine/net_ws.cpp
@@ -784,7 +784,11 @@ qboolean NET_LagPacket(qboolean newdata, netsrc_t sock, netadr_t *from, sizebuf_
 	}
 
 	curtime = realtime;
+#ifdef REHLDS_FIXES
+	if (newdata && data)
+#else
 	if (newdata)
+#endif
 	{
 		if (fakeloss.value != 0.0)
 		{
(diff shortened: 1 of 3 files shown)


LevShisterov:
Also, please fix calls to NET_LagPacket and R_StudioCalcBonePosition where 0 is passed as a value for the pointer.

In-line:
@LevShisterov Okey 👍

In-line:
I will create another PR to refactor calls little bit.

In-line:
@LevShisterov #445

PR #444, opened 2017-04-26
Fix out of bounds assign in Sys_InitGame

In-line:
Found with cppcheck

### DIFF
--- a/rehlds/engine/sys_dll2.cpp
+++ b/rehlds/engine/sys_dll2.cpp
@@ -524,14 +524,7 @@ int Sys_InitGame(char *lpOrgCmdLine, char *pBaseDir, void *pwnd, int bIsDedicate
 		ClientDLL_ActivateMouse();
 
 	char MessageText[512];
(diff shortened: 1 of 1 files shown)


rehlds/engine/sys_dll2.cpp
```diff
@@ -524,10 +524,14 @@ int Sys_InitGame(char *lpOrgCmdLine, char *pBaseDir, void *pwnd, int bIsDedicate
 		ClientDLL_ActivateMouse();
 
 	char MessageText[512];
+#ifdef REHLDS_FIXES
+	const char en_US[] = "en_US.UTF-8";
+#else
 	const char en_US[12];
 
 	Q_strcpy(en_US, "en_US.UTF-8");
 	en_US[16] = 0;
```
Sh1ft0x0EF: Just replace 16 with 12 here

In-line: @Sh1ft0x0EF I think autosize is better.

PR #445, opened 2017-04-26
Refactor calls

In-line:


rehlds/engine/net_ws.cpp
```diff
@@ -1023,14 +1023,14 @@ qboolean NET_QueuePacket(netsrc_t sock)
 
 	if (ret == -1 || ret == MAX_UDP_PACKET)
 	{
-		return NET_LagPacket(0, sock, 0, 0);
+		return NET_LagPacket(FALSE, sock, NULL, NULL);
```
WPMGPRoSToTeMa: Should be `nullptr`, not `NULL`.

In-line: @WPMGPRoSToTeMa Such changes should be made on a larger scale

LevShisterov:
Most of the code use NULL.

rehlds/engine/net_ws.cpp
WPMGPRoSToTeMa: @In-line but better change `0` directly to `nullptr` (skip `NULL` stage).

In-line: @WPMGPRoSToTeMa we can define NULL as nullptr

WPMGPRoSToTeMa: @In-line better to replace all of them in another refactor commit.

PR #446, opened 2017-04-27
Fix Sys_FloatTime in linux  (issue #441)

### LINKED ISSUE #441: Sys_FloatTime() isn't steady
Sys_FloatTime() isn't steady, it means that clock can jump backward and forward.
We can have situation, where second call to Sys_FloatTime() returns value less than the first.

In-line:
Windows implementation is already steady.

### DIFF
--- a/rehlds/engine/sys_dll.cpp
+++ b/rehlds/engine/sys_dll.cpp
@@ -625,9 +625,17 @@ double Sys_FloatTime(void)
 	if ( !bInitialized )
 	{
 		bInitialized = true;
+#ifdef REHLDS_FIXES
+		clock_gettime(CLOCK_MONOTONIC_RAW, &start_time);
+#else
 		clock_gettime(CLOCK_MONOTONIC, &start_time);
+#endif		
 	}
+#ifdef REHLDS_FIXES
+	clock_gettime(CLOCK_MONOTONIC_RAW, &now);
+#else
 	clock_gettime(CLOCK_MONOTONIC, &now);
+#endif
 	return (now.tv_sec - start_time.tv_sec) + now.tv_nsec * 0.000000001;
 }
 
--- a/rehlds/engine/sys_dll.cpp
+++ b/rehlds/engine/sys_dll.cpp
@@ -619,15 +619,15 @@ double EXT_FUNC Sys_FloatTime(void)
 double Sys_FloatTime(void)
 {
 	static struct timespec start_time;
-	static bool bInitialized;
+	static bool bInitialized = false;
 	struct timespec now;
 
 	if ( !bInitialized )
 	{
-		bInitialized = 1;
-		clock_gettime(1, &start_time);
+		bInitialized = true;
+		clock_gettime(CLOCK_MONOTONIC, &start_time);
 	}
-	clock_gettime(1, &now);
+	clock_gettime(CLOCK_MONOTONIC, &now);
 	return (now.tv_sec - start_time.tv_sec) + now.tv_nsec * 0.000000001;
 }
 


LevShisterov:
It is questionable. Monotonic raw time flow could differ from clients, because it is not trying to follow world-wide time.

In-line:
@LevShisterov Can you explain it for me ?

Windows implementation that use [QPC is steady](https://msdn.microsoft.com/en-us/library/windows/desktop/dn553408(v=vs.85).aspx) and don't follow NTP adjustments.
> QPC is typically the best method to use to time-stamp events and measure small time intervals that occur on the same system or virtual machine. Consider using GetSystemTimePreciseAsFileTime when you want to time-stamp events across multiple machines, provided that each machine is participating in a time synchronization scheme such as Network Time Protocol (NTP). QPC helps you avoid difficulties that can be encountered with other time measurement approaches, such as reading the processor’s time stamp counter (TSC) directly.

Sys_FloatTime() calculates time **from server start** 
`return (now.tv_sec - start_time.tv_sec) + now.tv_nsec * 0.000000001;`

LevShisterov:
Yes, QPC isn't adjusted by NTP. But whats the actual reason (or problems it solves) for the proposed patch?

WPMGPRoSToTeMa:
default from [here](https://www.reddit.com/r/programming/comments/y3mxv/clock_monotonic_avoiding_problems_with_leap/c5shfkg/).

We are using Sys_FloatTime not only for short and medium time periods, but also for long. So, that's why better to use `CLOCK_MONOTONIC`.

In-line:
@LevShisterov @WPMGPRoSToTeMa 

I think behavior of Linux and Windows versions should be same.
Actual issue that motivated me is in [HLTV](https://github.com/dreamstalker/rehlds/pull/440/commits/248a77b0f2e99e29a93d54764d01d2b48b5ee26b#diff-0e4264542704c61b16636c0189e79471R551)

- Expected behavior: Everything works fine in Linux and Windows
- Actual: Windows version works fine, Linux version floods console with warnings when time jumps backward.

I agree that code in HLTV should be changed too, but we need to specify exactly **one** behavior for Sys_FloatTime function.

WPMGPRoSToTeMa:
> Linux version floods console with warnings when time jumps backward

Have you succeeded to reproduce that?

In-line:
@WPMGPRoSToTeMa Yes.

WPMGPRoSToTeMa:
@In-line is delta 0 or negative?

WPMGPRoSToTeMa:
I'm also not sure about QPC, there is some workaround code for negative delta in windows version of `Sys_FloatTime`.

LevShisterov:
What you are both about, ReHLDS or ReHLTV?

In-line:
@WPMGPRoSToTeMa I didn't add logging. 
> QPC is independent of and isn't synchronized to any external time reference. To retrieve time stamps that can be synchronized to an external time reference, such as, Coordinated Universal Time (UTC) for use in high-resolution time-of-day measurements, use GetSystemTimePreciseAsFileTime.

> QPC is typically the best method to use to time-stamp events and measure small time intervals that occur on the same system or virtual machine.

[Source](https://msdn.microsoft.com/en-us/library/windows/desktop/dn553408(v=vs.85).aspx)

@LevShisterov I'm (at least) about ReHLDS, but actually `CCounter` should be changed too.

In-line:
Okey, I created a ~big~little confusion :smile: . We need to separate discussion about ReHLDS and ReHLTV. This is actually about Sys_FloatTime() steadiness, but ReHLTV uses QPC too and has Linux implementation that isn't steady.

WPMGPRoSToTeMa:
> @WPMGPRoSToTeMa I didn't add logging.

If it is 0 then it is potentially ok.

> Source

I mean there are some topics about QPC "jumps".

In-line:
@LevShisterov What do you say ?

LevShisterov:
I see no reason for ReHLDS to log something about time go backwards, because CLOCK_MONOTONIC is being used, which shouldn't jump backward.

In-line:
@LevShisterov I found some interesting links. Information is contradictory, but it seems, that old Linux versions implementations can jump backward and it's bug.
https://patchwork.kernel.org/patch/60563/
https://bugzilla.redhat.com/show_bug.cgi?id=448449
https://github.com/rust-lang/rust/issues/37902
http://stackoverflow.com/questions/3657289/linux-clock-gettimeclock-monotonic-strange-non-monotonic-behavior

I'll close PR for this moment.

LevShisterov:
Can you, please, instead just use first commit? It never hurts to use real defines.

LevShisterov:
Thanks!

PR #447, opened 2017-04-27
Add missing NOXREF and NOXREFCHECK to Host_Crash_f

In-line:


LevShisterov:
It should be in host_cmd.h too.

PR #457, opened 2017-05-05
HLTV: Fix crash and small refactoring.

s1lentq:
HLTV: Removed the obsolete filesystem_steam.

### DIFF
--- a/rehlds/HLTV/Core/src/World.cpp
+++ b/rehlds/HLTV/Core/src/World.cpp
@@ -1060,65 +1060,65 @@ void World::ParseClientData(BitBuffer *stream, unsigned int deltaSeqNr, BitBuffe
 	}
 }
 
-int World::ParseDeltaHeader(BitBuffer *stream, bool *remove, bool *custom, int *numbase, bool *newbl, int *newblindex, bool full, int *offset)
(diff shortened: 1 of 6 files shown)


rehlds/HLTV/Core/src/World.cpp
```diff
@@ -1060,65 +1060,65 @@ void World::ParseClientData(BitBuffer *stream, unsigned int deltaSeqNr, BitBuffe
...
-		else if (stream->ReadBit() == 0)
-		{
-			num = stream->ReadBits(6) + *numbase;
-		}
-		else
-		{
-			num = stream->ReadBits(11);
-		}
+		isdelta = stream->ReadBit() ? true : false;
+		remove = false;
```
In-line: "remove" is incorrectly highlighted in Github, mb rename it?

WPMGPRoSToTeMa: @In-line `isRemove`?

In-line: @WPMGPRoSToTeMa, I think it's good. I also emailed Github support about this.

IgnacioFDM: I don't think variable names should be decided based on github's syntax highlighting. No need to change it IMO

LevShisterov: +1 to the 

> variable names shouldn't be decided based on github's syntax highlighting

But variable still can be renamed.

PR #492, opened 2017-07-10
Use mtune=generic instead of march for GCC

In-line:


theAsmodai:
Yes, it will be optimal for local building.

PR #494, opened 2017-07-11
Fix accuracy, improve bullet physics. 

In-line:


stamepicmorg:
Самый полезный коммит, который я тут видел 💃

voed:
> ...linux version of HLDS, engine_i486.so

ReHLDS now affecting not only engine_i486.so, so i think this remark can be removed. It will give "стрельба как в 5787".

theAsmodai:
Impressive! You really found the problem.

WPMGPRoSToTeMa:
:trollface:

fred0r:
Reads nice.
Thank you so much for supporting the best FPS-Game ever !

Phantomsq:
Omfg. Finally!

IgnacioFDM:
We should probably increase API major version because of this.

justgo97:
nice catch.

PR #505, opened 2017-07-23
[UNDONE] Fix gcc compile

In-line:


In-line:
@s1lentq prepared better fix.

PR #513, opened 2017-08-01
[NEED TESTS AND REVIEWS] Reworked EXT_FUNC usage

WPMGPRoSToTeMa:
- Added missed EXT_FUNC (some engfuncs, interfaces, API)
- Removed redundant EXT_FUNC
- Removed EXT_FUNC from hookchain original functions since API users can't access them directly
- Moved EXT_FUNC to declaration
- Moved EXT_FUNC to function pointer access place via WRAP_EXT_FUNC
- Added EXT_FUNC to export macros
- Removed VGuiWrap2_IsInCareerMatch & VGuiWrap2_GetLocalizedStringLength duplicate declarations
- Moved some API functions to API file
- Hide public destructors (make them protected)


rehlds/engine/pr_cmds.h
```diff
@@ -126,32 +126,32 @@ void SV_SetGlobalTrace(trace_t *ptrace);
...
+EXT_FUNC edict_t *PF_Spawn_I(void);
 edict_t *CreateNamedEntity(int className);
 void PF_Remove_I(edict_t *ed);
 void PF_Remove_I_internal(edict_t *ed);
 edict_t *PF_find_Shared(int eStartSearchAfter, int iFieldToMatch, const char *szValueToFind);
 int iGetIndex(const char *pszField);
-edict_t *FindEntityByString(edict_t *pEdictStartSearchAfter, const char *pszField, const char *pszValue);
-int GetEntityIllum(edict_t *pEnt);
+EXT_FUNC edict_t *FindEntityByString(edict_t *pEdictStartSearchAfter, const char *pszField, const char *pszValue);
+EXT_FUNC int GetEntityIllum(edict_t *pEnt);
```
LevShisterov: Please use the same type/modifiers sorting. `EXT_FUNC int` vs `int EXT_FUNC`.

rehlds/rehlds/rehlds_api_impl.cpp
```diff
@@ -378,29 +252,91 @@ bool EXT_FUNC SV_EmitSound2_internal(edict_t *entity, IGameClient *pReceiver, in
...
 
+CRehldsServerStatic g_RehldsServerStatic;
+CRehldsServerData g_RehldsServerData;
+CRehldsHookchains g_RehldsHookchains;
+RehldsFuncs_t g_RehldsApiFuncs =
+{
+	&SV_DropClient_api,
+	&SV_RejectConnection,
+	&WRAP_EXT_FUNC(&Steam_NotifyBotConnect_api),
+	[]() EXT_FUNC { return &net_message; },
```
WPMGPRoSToTeMa: Looks like `EXT_FUNC` doesn't work for lambdas, but `EXT_FUNC` makes no sense for just returning pointers/values (but it can be potentially risky in debug mode).

LevShisterov:
For easier reviewing please split EXT_FUNC refactoring into a separate PR.
About WRAP_EXT_FUNC - will there be any performance loss?

Issue #31, opened 2015-06-08
Linux binaries are too large (>2MB)


dreamstalker:
Examine what the hell is going on with linux binaries - they are 2+ times larger than windows binaries

a1batross:
Can it be from b6c01cd ?

libstdc++ is a big library, why it need to be linked statically?

theAsmodai:
It's was done for compatibility with systems with older versions of glibc. As far as I observed, most of the increase was given by inlining functions without removing their body. Intel compiler very likes to inline everything that small or called from 1-2 places. Only jitasm.h gives about 400kb. MSVC generates for this about 3 times less code.

a1batross:
I think, better to look in Linux Base Standard, if you want to one Linux
binary for many distros. LSB defines minimal glibc and other version.

Also, why icc? There is a bunch of many open source and "good in
optimization" compilers: GCC and Clang.

21:50, ср, 08.07.2015, theAsmodai <notifications@github.com>:

theAsmodai:
ICC generates faster code than other compilers, and of course uses Intel's inline assembler syntax. It's used not frequently, but much nicer than AT&T syntax used in GAS.

a1batross:
I know about it.
On Linux general compiler is GCC, so I just wondering why it's not
supported. And software usually linked with shared libraries.

Ah, anyway, thanks for answer. :)

22:13, ср, 08.07.2015, theAsmodai <notifications@github.com>:

dreamstalker:
> Also, why icc? There is a bunch of many open source and "good in
> optimization" compilers: GCC and Clang.
Because it supports naked functions (i.e. without "push ebp; mov ebp, esp" prologue) and msvc syntax of inline assembly. Both features were crucial during the reverse engineering phase when we were installing tons of hooks into the original engine's code. 

Both points are not actual now, so GCC is unsupported only because of my laziness.


> And software usually linked with shared libraries.
When using most recent compiler you have to link some libraries statically to support conservative distros like CentOS

satnatantas:
>ICC generates faster code than other compilers

http://www.agner.org/optimize/blog/read.php?i=49#49
http://www.extremetech.com/computing/193480-intel-finally-agrees-to-pay-15-to-pentium-4-owners-over-amd-athlon-benchmarking-shenanigans

P.S. Or are  you patching your executables?

https://github.com/jimenezrick/patch-AuthenticAMD

theAsmodai:
[...]

PR #527, opened 2017-09-28
Optimize size by removing unused data and functions

In-line:
I tested this a little bit seems to not break AMX Mod X Hamsandwich module.
This reduced size of engine_i486.so by ~0.6 megabytes. 

Info about `--gc-sections`: https://stackoverflow.com/questions/6687630/how-to-remove-unused-c-c-symbols-with-gcc-and-ld
Info about version script: alliedmodders/amxmodx#447

Credits: @WPMGPRoSToTeMa, @s1lentq


theAsmodai:
Strange. I thought we already did it. Referred to #31

PR #568, opened 2018-01-13
Enable IPTOS_LOWDELAY by default

In-line:
https://en.wikipedia.org/wiki/Type_of_service
http://man7.org/linux/man-pages/man7/ip.7.html

### DIFF
--- a/rehlds/engine/net_ws.cpp
+++ b/rehlds/engine/net_ws.cpp
@@ -1555,7 +1555,11 @@ SOCKET NET_IPSocket(char *net_interface, int port, qboolean multicast)
 	}
 
 #ifndef _WIN32
+#ifdef REHLDS_FIXES
+	if (!COM_CheckParm("-notos"))
+#else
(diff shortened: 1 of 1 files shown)


WPMGPRoSToTeMa:
Can you write about the impact of this change?

In-line:
@WPMGPRoSToTeMa Yes. 

> Linux sends IPTOS_LOWDELAY datagrams first by
              default, but the exact behavior depends on the configured
              queueing discipline

Logically at least it will minimize server-client ping.

In-line:
@theAsmodai Is there any reason why you don't merge this?

theAsmodai:
I'm was offline one week. Are you sure that this change makes any sense?

In-line:
@theAsmodai Yes, it can improve user experience for players with high ping.

theAsmodai:
AFAIK it's outdated and useless flag. Routers doesn't handle it.

PR #569, opened 2018-01-14
[WIP] `-netthread` fixes and implementation for linux 

In-line:


In-line:
After several tests was confirmed that -netthread is not stable for now. Ping can increase 100+ at random time.

In-line:
It was tested by several people at least nothing bad was reported for the time going.

theAsmodai:
is it ready and tested?

In-line:
@theAsmodai Yes, it it ready and was tested by serveral people.

rehlds/engine/net_ws.cpp
```diff
@@ -1161,33 +1186,45 @@ DWORD WINAPI NET_ThreadMain(LPVOID lpThreadParameter)
 				break;
 		}
 
+#ifndef REHLDS_FIXES
 		Sys_Sleep(1);
```
theAsmodai: What is the purpose of this sleep? I see it will be caleed if no packets available while 0.25 sec.

rehlds/engine/net_ws.cpp
```diff
@@ -1198,11 +1235,38 @@ void NET_StopThread()
 	{
 		if (net_thread_initialized)
 		{
+#ifdef REHLDS_FIXES // Safe thread stopping.
+			NET_ThreadLock();
+			net_thread_terminated = true;
```
theAsmodai: You don't need lock here.

In-line:
You can think about this as poor man's `-pingboost 3`

theAsmodai:
I don't understand you.

In-line:
@theAsmodai Don't merge this for now. It seems that we need to place memory barriers. https://stackoverflow.com/questions/286629/what-is-a-memory-fence

theAsmodai:
Here it's not necessary.

rehlds/engine/net_ws.cpp
WPMGPRoSToTeMa: @theAsmodai what's the reason? x86?

PR #584, opened 2018-02-21
Default sv_rollangle/sv_rollspeed to 0.0

In-line:
This should'nt change behavior, because prior #583 cvars were not initialized.


In-line:
`REHLDS_FIXES` check in 0d5a8ee2bf6107b59a7bd1e82afde8853d8a7453 is redundant, so merging this anyway :smile:

PR #596, opened 2018-04-01
Fixes for new exploit involving " in info strings.

In-line:
Related to #595 .

### DIFF
--- a/rehlds/engine/sv_main.cpp
+++ b/rehlds/engine/sv_main.cpp
@@ -1977,29 +1977,29 @@ int SV_CheckForDuplicateSteamID(client_t *client)
 
 int SV_CheckForDuplicateNames(char *userinfo, qboolean bIsReconnecting, int nExcludeSlot)
 {
-	const char *val;
-	int i;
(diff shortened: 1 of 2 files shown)


In-line:
It was really pain to simplify Info_IsValid (removing redundant copies, copy-paste and so on). Behavior shouldn't be changed, so I didn't put refactoring under `REHLDS_FIXES`

WPMGPRoSToTeMa:
What about reverting #595?

rehlds/engine/sv_main.cpp
```diff
@@ -1979,6 +1979,7 @@ int SV_CheckForDuplicateNames(char *userinfo, qboolean bIsReconnecting, int nExc
 {
 	int dupc = 0;
 	char rawname[MAX_NAME];
+	char newname[MAX_NAME];
```
In-line: OMG. REALLY? BECAUSE OF THIS BUILD FAILED?

In-line:
@WPMGPRoSToTeMa Any suggestions?

PR #597, opened 2018-04-03
Dummy assignment?

IgnacioFDM:


In-line:
Maybe reversing mistake?

theAsmodai:
Yes, it's mistake. This is similar to the result of reverse of CMOVcc construction.

PR #604, opened 2018-04-16
Info code refactoring

theAsmodai:
I have made a lot of changes and hope all potencial bugs was fixed. Waiting for reviews.


WPMGPRoSToTeMa:
`sv_rehlds_userinfo_transmitted_fields` default value meaning changed. Can this break some games?

rehlds/engine/info.cpp
```diff
@@ -261,19 +409,49 @@ qboolean Info_IsKeyImportant(const char *key)
...
+	size_t largestLen = 0;
+
+	while (*s == '\\')
+	{
+		// skip starting slash
+		const char* key = ++s;
+
+		// skip key
+		while (*s != '\\')
+			s++;
```
LevShisterov: Can get past end of the string.

theAsmodai:
I'm added default keys used in Half-Life. Don't know any mods that using custom keys. If you think rehlds should transmit all fields by default, how would you suggest to implement it?

WPMGPRoSToTeMa:
> how would you suggest to implement it?

Something of these:
```
sv_rehlds_userinfo_transmitted_fields "all"
sv_rehlds_userinfo_transmitted_fields "compat"
sv_rehlds_userinfo_transmitted_fields "legacy"
sv_rehlds_userinfo_transmitted_fields "old"
sv_rehlds_userinfo_transmitted_fields "\..."
sv_rehlds_userinfo_transmitted_fields "\all..."
sv_rehlds_userinfo_transmitted_fields "\...all"
```

rehlds/public/strtools.h
```diff
@@ -153,8 +153,8 @@ char *Q_strlcpy(char (&dest)[size], const char *src) {
 }
 
 inline char *Q_strnlcpy(char *dest, const char *src, size_t n) {
-	Q_strncpy(dest, src, n - 1);
-	dest[n - 1] = '\0';
+	Q_strncpy(dest, src, n);
+	dest[n] = '\0';
```
LevShisterov: This is, probably, not a good idea, because function goes beoynd specified size. You can just add description to a function, that it is copy n-1 chars and adds a newline in the end.

LevShisterov:
`sv_rehlds_userinfo_transmitted_fields` - empty value should work as before. So in case mod needs some custom info fields, server operator can set empty value for the cvar.

theAsmodai:
I would like to add protection against wrong usage of sv_rehlds_userinfo_transmitted_fields. In my version empty value mean that only critical engine fields will be transmitted. So, cstrike owners can remove topcolor and bottomcolor since they are not used. Maybe we can add an additional cvar to on/off this option?

IgnacioFDM:
I'd recommend not breaking compatibility with sv_rehlds_userinfo_transmitted_fields

rehlds/engine/info.cpp
```diff
@@ -599,52 +964,92 @@ qboolean Info_IsValid(const char *s)
...
+				return x.integer;
+		}
+		return false;
+	};
+
+	for (char *key = Q_strtok(keys, "\\"); key; key = Q_strtok(nullptr, "\\"))
+	{
+		if (key[0] == '_') {
+			Con_Printf("%s: private keys couldn't be transmitted.\n", __FUNCTION__);
+			continue;
```
LevShisterov: It will spam in console each time. Would be good to output key name at least.

rehlds/engine/info.cpp
```diff
@@ -599,52 +964,92 @@ qboolean Info_IsValid(const char *s)
...
+		}
 
+		if (Q_strlen(key) >= MAX_KV_LEN) {
+			Con_Printf("%s: keys and values must be < %i characters\n", __FUNCTION__, MAX_KV_LEN);
+			continue;
```
LevShisterov: The same.

rehlds/engine/info.cpp
```diff
@@ -599,52 +964,92 @@ qboolean Info_IsValid(const char *s)
...
+		}
+	}
+}
+
+void Info_CollectFields(char *destInfo, const char *srcInfo, size_t maxsize)
+{
+	char add[512], valueBuf[32];
+	size_t userInfoLength = 0;
+
+	for (auto field : g_info_transmitted_fields)
```
LevShisterov: Can I suggest to rename it to `g_info_transmit_fields`?

LevShisterov: sv_rehlds_userinfo_transmitted_fields probably should be renamed too.

LevShisterov:
About `sv_rehlds_userinfo_transmitted_fields` cvar and `g_info_important_fields`, I have a word from a side that some server ops wish to block `*hltv` and `*sid` sending to hide steamids, etc. So, probably mark these as not important and include in default value?

rehlds/engine/info.cpp
```diff
@@ -97,24 +169,61 @@ const char* EXT_FUNC Info_ValueForKey(const char *s, const char *key)
...
+
+		size_t keyLen = s - key;
+		++s; // skip separating slash
+
+		// skip value
+		while (*s != '\\' && *s != '\0')
+			s++;
+
+		if (keyLen < lookupLen)
+			continue;
```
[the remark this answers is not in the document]
theAsmodai: I'm copied the default engine behaviour. But since duplicate keys will be checked, we can change this to exact match.

rehlds/engine/info.cpp
```diff
@@ -239,12 +379,28 @@ void Info_RemovePrefixedKeys(char *s, const char prefix)
...
 
+#ifdef REHLDS_FIXES
+qboolean Info_IsKeyImportant(const char *key, size_t keyLen)
+{
+	if (key[0] == '*')
+		return true;
+
+	for (auto& field : g_info_important_fields) {
+		if (!Q_strncmp(key, field.name, keyLen))
+			return true;
```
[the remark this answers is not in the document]
theAsmodai: Yes, you are right. I'm forgot to recheck it.

LevShisterov:
> *sid blocking will break avatars showing on clients.

Yes, this is the desired (not by me) effect.

> anyway hiding of steamid is a bad idea because it was designed as public identificator of players.

I am also not welcome the idea of blocking `status` and other steamid sources. But, still, we are open source, why not let people to shoot in the leg? Anyway, it's up to you, I don't care about this.

rehlds/engine/info.cpp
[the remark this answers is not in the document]
LevShisterov: You are right, it was so before. I think we should change this.

Phantomsq:
Is it going to be merged?

rehlds/engine/info.cpp
```diff
@@ -541,55 +836,133 @@ void Info_Print(const char *s)
...
+		size_t valueLen = s - value;
+		if (valueLen == 0 || valueLen >= MAX_KV_LEN)
+			return FALSE;
+
+		if (*s == '\0')
+			return TRUE;
+
+		existingKeys[existingKeysNum].start = key;
+		existingKeys[existingKeysNum].len = keyLen;
+		existingKeysNum++;
```
LevShisterov: Ah, one more thing: you should somehow assure that you will not go past `existingKeys` border. I think it should be incoming buffer length checking at the function start.

theAsmodai: Ok. And last question: should `sv_rehlds_userinfo_transmitted_fields` be renamed or not? I'm created wiki page https://github.com/dreamstalker/rehlds/wiki/Userinfo-keys and later I will will set the correct default value and description for this cvar.

LevShisterov: Thought about this, probably non-default value is used often in CS, so we can leave it as it is.

theAsmodai:
Tests was fixed.

Issue #17, opened 2015-05-20
Bits writing MSG_ functions may write data outside allocated buffer when sizebuf_t is overflowed


dreamstalker:
Bits writing MSG_ functions (MSG_WriteBits, MSG_WriteOneBit) are designed in a way that the byte being currently written lays outside of cursize of sizebuf on which bits writing is performed.
This could be clearly seen in MSG_StartBitWriting's code
<pre>
	bfwrite.nCurOutputBit = 0;
	bfwrite.pbuf = buf;
	bfwrite.pOutByte = &buf->data[buf->cursize]; // <<<<=========
</pre>
Here we set the bfwrite.pOutByte to a pointer in the buf->data buffer that lays outside buf->curside, and, possibly, may even lay outside allocated buf->data buffer (when cursize == maxsize). Notice that SZ_GetSpace() is not called here, so SIZEBUF_OVERFLOWED flag will not be set.

If we then call MSG_WriteOneBit(), data will be written into bfwrite.pOutByte pointer, without calling SZ_GetSpace(). In that case 1 byte after the allocated buf->data buffer will be corrupted.

Another piece of crappy code is hiding inside MSG_WriteBits:
<pre>
	int surplusBytes = 0;
	if ((uint32)bfwrite.nCurOutputBit >= 8)	{
		surplusBytes = 1; //surplusBytes could be either 0 or 1
		....
	}
	....
		SZ_GetSpace(bfwrite.pbuf, surplusBytes + 4); //allocate 4 or 5 bytes
		if (!(bfwrite.pbuf->flags & SIZEBUF_OVERFLOWED))
		{
			// write 4 bytes (it's ok)
			*(uint32 *)bfwrite.pOutByte = (data << bfwrite.nCurOutputBit) | *(uint32 *)bfwrite.pOutByte & ROWBITTABLE[bfwrite.nCurOutputBit];
			int leftBits = 32 - bfwrite.nCurOutputBit;
			bfwrite.nCurOutputBit = bits & 7;
			bfwrite.pOutByte += 4;
			*(uint32 *)bfwrite.pOutByte = data >> leftBits; //then write another 4 bytes! wtf?
		}
</pre>
Here (in some rare cases) we allocate 4 or 5 bytes are from sizebuf and then write 8 bytes there. If sizebuf's data is the last variable allocated on stack and -fomit-frame-pointers is used, the whole return address will be rewritten and hlds will experience a (not) funny trip into nowhere when it executes the ret instruction, he-he.

PR #609, opened 2018-04-30
fix for split packets

verimavcisi:
its a fix for split attack
server dont have to receive split packets but hlds can handle it
when someone spams split packets, server's buffer will be filled

open source split spammer:
https://raw.githubusercontent.com/verimavcisi/projeler/master/linux/half-life/split_spammer/split_spammer.c
compile: gcc split.spammer.c
usage: sudo ./a.out sourceip destinationip destinationport packetdelay

summary when you start split spammer:
not connectable server WITHOUT this commit
no ping difference or no any lag WITH this commit

### DIFF
--- a/rehlds/engine/net_ws.cpp
+++ b/rehlds/engine/net_ws.cpp
@@ -776,14 +776,6 @@ qboolean NET_GetLong(unsigned char *pData, int size, int *outSize)
 		gNetSplit.currentSequence = pHeader->sequenceNumber;
 		gNetSplit.splitCount = packetCount;
 
-#ifdef REHLDS_FIXES
-		gNetSplit.totalSize = 0;
-
-		// clear part's sequence
-		for (int i = 0; i < NET_WS_MAX_FRAGMENTS; i++)
-			gNetSplitFlags[i] = -1;
-#endif
-
 		if (net_showpackets.value == 4.0f)
 			Con_Printf("<-- Split packet restart %i count %i seq\n", gNetSplit.splitCount, sequenceNumber);
 	}
@@ -809,9 +801,6 @@ qboolean NET_GetLong(unsigned char *pData, int size, int *outSize)
 		if (SPLIT_SIZE * packetNumber + packetPayloadSize > MAX_UDP_PACKET)
 		{
 			Con_NetPrintf("Malformed packet size (%i, %i)\n", SPLIT_SIZE * packetNumber, packetPayloadSize);
-#ifdef REHLDS_FIXES
-			gNetSplit.currentSequence = -1;
-#endif
 			return FALSE;
 		}
 
@@ -833,9 +822,6 @@ qboolean NET_GetLong(unsigned char *pData, int size, int *outSize)
 					i + 1,
 					gNetSplitFlags[i],
 					gNetSplit.currentSequence);
-#ifdef REHLDS_FIXES
-				gNetSplit.currentSequence = -1; // no more parts can be attached, clear it
-#endif
 				return FALSE;
 			}
 		}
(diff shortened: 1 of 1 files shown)


theAsmodai:
Hello. I think you don't understand how it works. This flooder sends 4-byte fffffffe header from spoofed addresses which can cause only flood to the console in developer mode. You should not use developer mode on public servers.

theAsmodai:
Your experiment is incorrect because testing client was launched on the same machine where the flooder. I think it is the only reason why do you see lags.

krangm:
actually i helped @verimavcisi to test that spammer on different servers im hosting.
it ealisy makes server unplayable

justgo97:
yeah this split packet flood can make server unplayable and i guess just a cvar for turning the console message OFF/ON is enough

verimavcisi:
theAsmodai
i tested not local rehlds servers too (with precompiled binary and my commit)

krangm
thanks for reply

justgo97
you watchted video what i recorded ?
i cant see any console message when i spam splits but server gets lag

verimavcisi:
theAsmodai
this code only working when packet is split and packet size lower than 9
`
Con_NetPrintf("Invalid split packet length %i\n", in_message.cursize);
return FALSE;
`
net_ws.cpp NET_QueuePacket's this part just checking split packet's accuracy
"
https://i.imgur.com/3tCIfSh.png
if everything is okay for split packet call NET_GetLong function
"
problem is when you call NET_GetLong every 15625< (default 64 splits in second) micro second server's buffer will be filled
my commit returns instantly NET_QueuePacket function when packet type is split
i never see hl client send split packet for anything so split should be not exist

verimavcisi:
@WPMGPRoSToTeMa i dont think so
if this code really blocking split packets why precompiled rehlds binary lagging ? (with my commit server dont get lags)

WPMGPRoSToTeMa:
@verimavcisi try commenting `Con_NetPrintf` and test again.

verimavcisi:
@WPMGPRoSToTeMa server still gets lag when i open split spammer

WPMGPRoSToTeMa:
@verimavcisi try replacing `return FALSE;` to `return TRUE;`.

verimavcisi:
@WPMGPRoSToTeMa tested on local and not local server, only no lag for local server, this line should be not exist maybe ? if (sock == NS_SERVER)

bye the way code going to be same thing with my pull request

theAsmodai:
I think you enabled sv_cheats and fakelag cvars. Theay are should be 0.

verimavcisi:
@theAsmodai man, you should try it by yourself instead of blame me with useless messages

theAsmodai:
No, you should justify the need of suggested changes or I close this useless thread.

Shigbeard:
Justification: Without this, my server was lagged out so hard the server's watchdog couldn't verify it was even up. With this patch made by @verimavcisi, the attack that he attempted on my server, the one I attempted on my server, and the one my other server attempted on my server, all failed to impact gameplay or even the server watchdog.

Note that in my tests, sv_cheats was disabled. There was no fakelag in use. Before implementing this patch, I was using the latest version of rehlds and the only change that occured between stock hlds and rehlds was that rehlds suppressed an error message in the server console.

The point I'm trying to make @theAsmodai is that rehlds DOES NOT currently protect against specially crafted split packet spam. This pull request implements changes to rehlds that DOES protect against specially crafted split packet spam. I've tested it first hand. @verimavcisi tested it with me. I've tried to use split packet spamming from 3 different sources: Internally from same machine as server, externally from same IP as a connected client, and externally from an IP that is neither connected to the server as a client or even on the same subnet as the client. All of these sources could virtually halt the server before installing this patch. All of these sources could not affect the server after installing this patch.

There's your justification for this pull request. 
✋ 
💨 
🎤

theAsmodai:
> The point I'm trying to make @theAsmodai is that rehlds DOES NOT currently protect against specially crafted split packet spam.

You are wrong. **ALL** split packets already disabled without any error messages, logs, etc... I'm close this PR because anyway it's incorrect. If you disagree you can investigate rehlds sources deeper and describe founded problem in issues section. I'm waiting for something like this: https://github.com/dreamstalker/rehlds/issues/17
If the problem really exists, then we will think how to fix it.

LevShisterov:
With Con_DPrintf there is some overwork with flight recording.

PR #624, opened 2018-06-07
Theoretically better unlag based on bose position restoring

a1batross:
How it (should) work?

The main idea is to remember client's bone position on each frame. It's intended to help with situation where one player is updated in real-time, but lagging second player not yet got this updates. Unlag fixes it only for player position(i.e. working when player is running) and does not restores any other states.

Example: 
Let's say we have a server with two players and they "see" each other at one frame lag.
* Frame[0]: 
  * Player[0]:
    * Physics is running.

* Frame[1]: 
  * Player[0]:
    * Physics is running.
  * Player[1]:
    * Player[1] is connected.
    * Physics is running.

* Frame[2]:
  * Player[0]:
    * Player is ducking, i.e. his head is a bit below.
    * Physics is running. 
  * Player[1]:
    * Player[1] shoots at Player[0] right to his head. Due to lag, Player[1] sees Player[0] as NOT ducking.
    * Physics is running. Bullet is traced.
    * WITH bone position history:
       * Player[0] origin restored at Frame[1] **AND** head hull position is restored at Frame[1] position.
    * WITHOUT bone position history:
       * Player[0] origin restored at Frame[1], **BUT** his bones are set up in Frame[2] position, because previous sequence is unknown.

Yes, this demonstrates only sequence changing(i.e. ducking) but due to unlag imperfection, there can be  more real life examples.

P.S. I did not tested this at all as I don't have any server and players. :)
P. P. S. Originally I thought to make this for Xash3D, but it would be faster to check this concept on bigger community.

### DIFF
--- a/rehlds/engine/sv_user.cpp
+++ b/rehlds/engine/sv_user.cpp
@@ -1012,12 +1012,12 @@ void SV_RunCmd(usercmd_t *ucmd, int random_seed)
 	gEntityInterface.pfnPlayerPostThink(sv_player);
 	gEntityInterface.pfnCmdEnd(sv_player);
 	
+	if (!host_client->fakeclient)
+		SV_RestoreMove(host_client);
+	
 #ifdef REHLDS_FIXES
 	SV_SaveBoneState( host_client, sv_player );
 #endif
-
-	if (!host_client->fakeclient)
-		SV_RestoreMove(host_client);
 }
 
 int SV_ValidateClientCommand(char *pszCommand)
@@ -2037,6 +2037,22 @@ void SV_SaveBoneState(client_t *_host_client, const edict_t *edict)
 	// get last outgoing frame
 	frame = &_host_client->frames[SV_UPDATE_MASK & (_host_client->netchan.outgoing_sequence)];
 	
+	// not a studio model
+	if( g_psv.models[i]->type != mod_studio )
+	{
+		frame->bonestate.valid = false;
+		return;
+	}
+	
+	studiohdr_t *hdr = (studiohdr_t*)Mod_Extradata(g_psv.models[edict->v.modelindex]);
+	
+	// shouldn't really happen
+	if( !hdr )
+	{
+		frame->bonestate.valid = false;
+		return;
+	}
+	
 	// set up bones
 	g_pSvBlendingAPI->SV_StudioSetupBones(
 		g_psv.models[edict->v.modelindex],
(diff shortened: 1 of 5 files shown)


a1batross:
Also, don't forget to set up sv_bone_unlag 1 in game config.

rehlds/engine/sv_user.cpp
```diff
@@ -1910,3 +2022,80 @@ void SV_FullUpdate_f(void)
...
+	extern float rotationmatrix[3][4]; // in r_studio.cpp
+
+	if( !SV_IsPlayerIndex( num )) // just in case
+		return;
+	
+	// get last outgoing frame
+	frame = &_host_client->frames[SV_UPDATE_MASK & (_host_client->netchan.outgoing_sequence)];
+	
+	// not a studio model
+	if( g_psv.models[i]->type != mod_studio )
```
WPMGPRoSToTeMa: `i` -> `edict->v.modelindex`

a1batross: Oops. Thanks :)

rehlds/engine/sv_user.cpp
```diff
@@ -1152,6 +1159,89 @@ entity_state_t *SV_FindEntInPack(int index, packet_entities_t *pack)
 	return NULL;
 }
 
+#ifdef REHLDS_FIXES
+void VectorsAngles( const vec3_t forward, const vec3_t right, const vec3_t up, vec3_t angles )
```
In-line: No need to include this under `REHLDS_FIXES`

di57inct:
is this the same as how source handles it? if so, PLEASE DONT FUCKING TOUCH IT
i don't want that cancerous source unlag on my goldsource games

rehlds/engine/sv_user.cpp
a1batross: @In-line ok, will remove in next commit.

a1batross:
@di57inct I don't know how Source doing it.

di57inct:
Sounds to me like that's it.
https://youtube.com/watch?v=6EwaW2iz4iA

a1batross:
@di57inct I hear only common words, without the details. Will the reasoning be better? Or just childish "it's cancerous"?

di57inct:
I don't have the patience to argue or explain it to you.
Don't force it into the main ReHLDS or if you do, make it optional.
Don't ruin GoldSrc games.

WPMGPRoSToTeMa:
> Sounds to me like that's it.

That's how it currently works (without this commit).
This commit adds only correct interpolation of animations, not player position, it'll be the same.

di57inct:
Okay I sure hope so. Thanks for the info.

2020karaulov2020:
This is  not fix for bad hitboxes ?))

a1batross:
Broken CS1.6 hitboxes? Of course not. It only makes them correct for small
time window when the animation switches.

2ReHLDS devs: I don't care about this patch anymore. Feel free to close it
or rework it without attribution.

пт, 5 февр. 2021 г., 19:26 Very Strange Karaulov <notifications@github.com>:

SNMetamorph:
Is it still actual? It'll be cool to implement it, because there are issues with hitreg in fast-paced mods

a1batross:
@SNMetamorph nah, not really.

You can try reimplement it for Xash though.

UnrealKaraulov:
No video with how to it works? No possibility to merge it?)

PR #644, opened 2018-08-12
Fix family of DoS attacks/exploits [help needed]

IgnacioFDM:
This fixes a family of DoS exploits. Old behavior was to stop processing packets altogether when you received something weird. With this commit, it will simply skip the erroneous packets and keep on processing.

This fixes, among other exploits, the following:
```Someone tries to send split packet to the server``` (0xFE exploit)
```Malformed packet number```

Help needed with the following:

- RCON testing needed
- Fakelag is untested and might be broken
- Linux is untested
- Threaded networking is untested
- Should we stop or skip on MAX_UDP_PACKET?

### DIFF
--- a/rehlds/engine/net.h
+++ b/rehlds/engine/net.h
@@ -414,3 +414,13 @@ typedef struct netchan_s
 #else // REHLDS_FIXES
 #define Con_NetPrintf Con_Printf
 #endif // REHLDS_FIXES
+
+#ifdef REHLDS_FIXES
+#define NET_SKIP -1
+#define NET_STOP 0
+#define NET_PROCESS 1
+#else // REHLDS_FIXES
+#define NET_SKIP FALSE
+#define NET_STOP FALSE
+#define NET_PROCESS TRUE
+#endif // REHLDS_FIXES
--- a/rehlds/engine/net_ws.cpp
+++ b/rehlds/engine/net_ws.cpp
@@ -672,7 +672,7 @@ void NET_AdjustLag()
 	}
 }
 
-qboolean NET_LagPacket(qboolean newdata, netsrc_t sock, netadr_t *from, sizebuf_t *data)
+int NET_LagPacket(int newdata, netsrc_t sock, netadr_t *from, sizebuf_t *data)
 {
 	if (gFakeLag <= 0.0)
 	{
@@ -755,7 +755,7 @@ void NET_FlushSocket(netsrc_t sock)
 	}
 }
 
-qboolean NET_GetLong(unsigned char *pData, int size, int *outSize)
+int NET_GetLong(unsigned char *pData, int size, int *outSize)
 {
 	static int gNetSplitFlags[NET_WS_MAX_FRAGMENTS];
 	SPLITPACKET *pHeader = (SPLITPACKET *) pData;
@@ -768,7 +768,7 @@ qboolean NET_GetLong(unsigned char *pData, int size, int *outSize)
 	if (packetNumber >= NET_WS_MAX_FRAGMENTS || packetCount > NET_WS_MAX_FRAGMENTS)
 	{
 		Con_NetPrintf("Malformed packet number (%i/%i)\n", packetNumber + 1, packetCount);
-		return FALSE;
+		return NET_SKIP;
 	}
 
 	if (gNetSplit.currentSequence == -1 || sequenceNumber != gNetSplit.currentSequence)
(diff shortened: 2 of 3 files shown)


OsweRRR:
it seems to be the same as okapi in hlds

`public Con_Printf_Hook(input[])
{
	if(containi(input, "Invalid split packet length") != -1) // malformed etc etc etc
	{
		return okapi_ret_supercede
	}
	return okapi_ret_ignore
}`

IgnacioFDM:
@OsweRRR 

Well, I've haven't seen that code or know really how okapi works, but I would guess from that snippet that it makes the packets return TRUE instead of FALSE, so they go through, which could lead to weird bugs, and old packets being replayed, assuming they aren't filtered at a later stage. Although if you were to ban them right there, and let the packet go through, it will get stopped at SV_FilterPacket.

This should be a much cleaner solution.

WPMGPRoSToTeMa:
Can you explain how does this help against split packet flood? (0xFFFFFFFE)
As I can see your PR just removes the `NET_LagPacket` call for that case, but `NET_LagPacket` is like NOOP.

IgnacioFDM:
>As I can see your PR just removes the NET_LagPacket call for that case

Where? 

The issue related to this family of DoS attacks is stopping SV_ReadPackets loop, when there is actually additional stuff to process (the legitimate traffic). I simply make it so that instead of exiting the loop, it continues processing the legitimate stuff.

WPMGPRoSToTeMa:
@IgnacioFDM ah, sorry. I misunderstood the description.

> Where?

Current code:
https://github.com/dreamstalker/rehlds/blob/0ce0af7d0f131058cc3e8e1317af1a5c53453903/rehlds/engine/net_ws.cpp#L938-L945
https://github.com/dreamstalker/rehlds/blob/0ce0af7d0f131058cc3e8e1317af1a5c53453903/rehlds/engine/net_ws.cpp#L1198-L1200
With your PR:
https://github.com/dreamstalker/rehlds/blob/17697cda5119a7dea896a36cb4e4dabb8b4d8bf1/rehlds/engine/net_ws.cpp#L938-L945
https://github.com/dreamstalker/rehlds/blob/17697cda5119a7dea896a36cb4e4dabb8b4d8bf1/rehlds/engine/net_ws.cpp#L1198-L1200

IgnacioFDM:
Oh I understand your point now, that I used ```if (bret == NET_STOP) ``` instead of ```if (bret == NET_STOP || bret == NET_SKIP) ```

This wasn't really intentional I believe, maybe ```if (bret == NET_STOP || bret == NET_SKIP) ``` would be more correct. I didn't really bother with that part because I wanted to push a fix fast and wasn't really interested with the fakelag part, and as you said it doesn't matter if fakelag is off.

Also, in NET_LagPacket, it's still needed to change TRUE/FALSE definitions to NET_SKIP, NET_STOP and NET_PROCESS. I'm quite busy at the moment and didn't have time to study how fakelag worked.

Please feel free to fix these things.

WPMGPRoSToTeMa:
I have looked into fakelag code and now I can say that `if (bret == NET_STOP)` is correct. But I think it would be better to slightly rework `NET_GetPacket` and `NET_QueuePacket` functions. Also I can't find fakelag code for split packets in original code, that's strange. I'll make another PR, but don't close this.

IgnacioFDM:
Did you check the MAX_UDP_PACKET case? I'm pretty sure it should be NET_SKIP.

WPMGPRoSToTeMa:
https://gist.github.com/WPMGPRoSToTeMa/2812ffef476802d05f0aee8a5149fb66, I'll make PR later.

RauliTop:
@WPMGPRoSToTeMa two weeks later...

Zetz512:
I download the commit with this changes and right now works great, the exploit can't crash the server 👍

bobito1993:
This is exploit program not fixet testing nev rehlds version :@ "malformet packet" :@
ReHLDS Exploit!

http://www.img-share.eu/f/images/540/Untitledg2WAVAyPQ3w354q3nwHS1532777368.png

bobito1993:
Test this exploit work not fixet :@ :@

http://alfa.kachi-snimka.info/vij.php?id=bwy1536191204u.png

IgnacioFDM:
@bobito1993 
It crashes because this fix has not been merged into master branch yet. 

Try downloading from here

http://teamcity.rehlds.org/viewLog.html?buildId=15162&buildTypeId=Rehlds_Publish&tab=artifacts#!-5adylxjtipds,-1dmn32mdeksv3,a94kejwrnjev,-1u2h3up2ep0wr


Also I've just arrived from vacations, and still have to do some work not related to hlds. It'd be nice if @WPMGPRoSToTeMa created a PR for his gist, and @theAsmodai , @s1lentq  and others reviewed either @WPMGPRoSToTeMa PR or my PR.

bobito1993:
Finally, the fix is tested I do not already work Fucking exploit of the new rehlds rehlds-dist-3.4.0.662-dee Thank you brother and the success of the team wonderful You are always so good from the bad Always have the good to remember the miserable people who make dirt

guiledu:
With this version rehlds-dist-3.4.0.662 amxmodx does not work?

WPMGPRoSToTeMa:
#652

PR #656, opened 2018-09-24
Fix named constants usages

WPMGPRoSToTeMa:


rehlds/public/rehlds/model.h
```diff
@@ -204,6 +204,7 @@ typedef struct hull_s
 	int				lastclipnode;
 	vec3_t			clip_mins, clip_maxs;
 } hull_t;
+typedef enum { point_hull = 0, human_hull = 1, large_hull = 2, head_hull = 3 };
```
WPMGPRoSToTeMa: Is it ok to add it here?

LevShisterov: Please change these to start from the hull_, i.e. hull_point.

rehlds/engine/pmovetst.cpp
```diff
@@ -145,14 +145,14 @@ int PM_LinkContents(vec_t *p, int *pIndex)
 		test[0] = p[0] - pe->origin[0];
 		test[1] = p[1] - pe->origin[1];
 		test[2] = p[2] - pe->origin[2];
-		if (PM_HullPointContents(model->hulls, model->hulls[0].firstclipnode, test) != -1) {
+		if (PM_HullPointContents(model->hulls, model->hulls[0].firstclipnode, test) != CONTENTS_EMPTY) {
```
Mistrick: Missed point name?

rehlds/public/rehlds/model.h
WPMGPRoSToTeMa: @LevShisterov this naming is used in the original Half-Life source code, moreover rehlds has this enum in `dlls/util.h`. The problem is: I can't just include `dlls/util.h` in `precompiled.h`, it causes compilation errors due to `Vector`. Maybe we should move this enum from `dlls/util.h` to another header?
Also we need another hull enum for `usehull`, `player_mins`/`player_maxs` and `GetHullBounds`, because it has different numeration. Do you have any ideas about it? About its naming?

rehlds/engine/pmovetst.cpp
WPMGPRoSToTeMa: @Mistrick thanks!

rehlds/public/rehlds/model.h
LevShisterov: I see. Ok. Probably model.h is a good place for it.

Not sure about hull numbers. It is a mess with them. For example in Mod_LoadClipnodes and in HLSDK HUD_GetHullBounds same indexes use different values.

LevShisterov:
Ye, I suppose that we can solve that if and when that will happen.

2018-09-27 21:26 GMT+05:00 Artem Golubikhin <notifications@github.com>:

WPMGPRoSToTeMa:
@LevShisterov currently I'm thinking about the names for constants of different hull numeration. I have several ideas for it:
1. `PMOVE_HULL_*` because it is related to playermove code.
2. `PLAYER_HULL_*` because it is related to `player_mins` and `player_maxs` variables.
3. `MOD_HULL_*`/`DLL_HULL_*`/`GAME_HULL_*`/something else because sizes for these hulls are controlled by gamedll via `GetHullBounds` callback.

Also I'm using `player_mins` and `player_maxs` in `SV_HullForBsp`:https://github.com/dreamstalker/rehlds/blob/65c6ce593b5eabf13e92b03352e4b429d0d797b0/rehlds/engine/world.cpp#L198-L208 So, now it is not only for players and looks like we should prefer the 3rd option. What do you think?

WPMGPRoSToTeMa:
@LevShisterov ping.

LevShisterov:
About hull enumeration.
Mod_LoadClipnodes makes the things crazy, it doesn't match player_mins/player_maxs. I mean that if you want to make enum for model.hulls all references to that field should be processed. And it is hard to determine exact namings for constants. Naming them HULL_1, HULL_2, etc doesn't makes a lot of sense. At least I wasn't able to figure out namings for these fast. Probably you have some suggestions.
In SV_GetPlayerHulls looks like MAX_MAP_HULLS should be used.

PR #660, opened 2018-09-26
Add hookchain for connectionless limiter

theAsmodai:


rehlds/engine/sv_main.cpp
```diff
@@ -3599,7 +3599,7 @@ void SV_ReadPackets(void)
 		if (*(uint32 *)net_message.data == 0xFFFFFFFF)
 		{
 			// Connectionless packet
-			if (SV_CheckConnectionLessRateLimits(net_from))
+			if (g_RehldsHookchains.m_SV_CheckConnectionLessRateLimits.callChain([](netadr_t& net_from, const uint8_t *, int) { return SV_CheckConnectionLessRateLimits(net_from); }, net_from, net_message.data, net_message.cursize))
```
WPMGPRoSToTeMa: NOTE: it is ok to use lambda without `EXT_FUNC` here, because `callOriginal` is already marked as `EXT_FUNC`.

theAsmodai: Yes, because it's internal rehlds function not exported in api.

Issue #685, opened 2019-02-26
TFC Server Problems


DarthMan:
Hello. After a few map changes, everytime I conenct to my TFC ReHLDS server I'm getting this error:

BUILD 1631 SERVER (0 CRC)
Server # 33
Host_Error: UserMsg: Not Present on Client 192

Before using ReHLDS, I didn't have such issues but the server was very crashy, whereas now, with ReHLDS, it's very rare that it crashes but, instead, I'm getting that error.
My server is Steam-only so the issue is not related to ReUnion or whatever.

DarthMan:
Problem was related to a plug-in. Case solved :-)

DarthMan:
Problem is bot solved, it is related to ReHLDS. With metamod disabled, I still have this issue after about 20 server restarts. With normal HLDS the server runs fine. But I really want to use ReHLDS.

DarthMan:
Here the full error: 

Last 32 messages parsed.
224912 6522 svc_newusermsg
224912 6541 svc_newusermsg
224912 6560 svc_newusermsg
224912 6579 svc_newusermsg
224912 6598 svc_newusermsg
224912 6617 svc_newusermsg
224912 6636 svc_newusermsg
224912 6655 svc_newusermsg
224912 6674 svc_newusermsg
224912 6693 svc_newusermsg
224912 6712 svc_newusermsg
224912 6731 svc_newusermsg
224912 6750 svc_newusermsg
224912 6769 svc_newusermsg
224912 6788 svc_newusermsg
224912 6807 svc_newusermsg
224912 6826 svc_newusermsg
224912 6845 svc_newusermsg
224912 6864 svc_newusermsg
224912 6883 svc_newusermsg
224912 6902 svc_newusermsg
224912 6921 svc_newusermsg
224912 6940 svc_newusermsg
224912 6959 svc_stufftext
224912 6992 svc_updateuserinfo
224974 0000 svc_resourcerequest
224974 0009 svc_resourcelocation
224974 0042 svc_resourcelist
224984 0000 svc_setview
225004 0000 svc_spawnstaticsound
225004 0015 svc_spawnbaseline
BAD:  4361:svc_bad
Host_Error: CL_ParseServerMessage: Illegible server message - svc_bad

Issue #606, opened 2018-04-23
ReHLDS Incompatability with TFC


MaZDuR:
Hello, I previously opened #423 last year, this bug is still present as recently as build/commit 641, I unfortunately don't have any more information other than this issue is only present under ReHLDS after a large period of time elapses.

SERVER
```status
hostname:  -[EVIL]- ReHLDS (the pinnacle of broken)
version :  48/1.1.2.2/Stdio 1455 secure  (20)
tcp/ip  :  66.151.244.170:28888
map     :  casbah at: 0 x, 0 y, 0 z
players :  0 active (10 max)

#      name userid uniqueid frag time ping loss adr
0 users
stats
CPU   In    Out   Uptime  Users   FPS    Players
 2.00  0.00  0.00    3197     9   90.48       0
version
Protocol version 48
Exe version 1.1.2.2/Stdio (tfc)
ReHLDS version: 3.4.0.641-dev
Build date: 16:11:38 Apr  6 2018 (1455)
Build from: https://github.com/dreamstalker/rehlds/commit/ce163a3
meta version
Metamod v1.21p38  2018/02/11 (5:13)
by Will Day
   http://www.metamod.org/
 Patch: Metamod-P (mm-p) v38
 by Jussi Kivilinna
    http://metamod-p.sourceforge.net/
compiled: Feb 11 2018, 11:05:06 EET (optimized)
meta list
Currently loaded plugins:
      description      stat pend  file              vers      src   load  unlod
0 plugins, 0 running
```

CLIENT
```Connecting to 66.151.244.170:28888...
Connection accepted by 66.151.244.170:28888

BUILD 1455 SERVER (0 CRC)
Server # 102
Serverinfo packet received.
Clearing memory
loading sprites/640hud3.spr
loading sprites/640hud7.spr
loading sprites/640hud8.spr
loading sprites/640hud9.spr
loading sprites/tfchud05.spr
loading sprites/tfchud01.spr
loading sprites/tfchud06.spr
loading sprites/640hud5.spr
loading sprites/640hud6.spr
loading sprites/crosshairs.spr
loading sprites/640hud4.spr
loading sprites/tfc_dmsg.spr
loading sprites/640hud1.spr
loading sprites/640hud2.spr
loading sprites/melee.spr
loading sprites/tfchud03.spr
loading sprites/tfchud04.spr
loading sprites/sniperrifle.spr
loading sprites/tfchud02.spr
loading sprites/autorifle.spr
[...]

Issue #423, opened 2017-04-05
"Illegible Server Message" after running for 12 hours


MaZDuR:
Hello, I am running into a problem where ReHLDS will disallow all clients from connecting and spit out an 'Illegible Server Message' to clients who try to connect, the server remains 'stable' and does not crash from this but requires a restart for this issue to be fixed.

meta list
```
Currently loaded plugins:
      description      stat pend  file              vers      src   load  unlod
 [ 1] AMX Mod X        open  -    amxmodx_mm.dll    v1.8.3-d  ini   Start ANY  
 [ 2] sparky utils v2  RUN   -    sparky_mm.dll     v2.02.1   ini   Start Never
2 plugins, 1 running
```
version
```
Protocol version 48
Exe version 1.1.2.2 (tfc)
ReHLDS version: 3.0.0.504-dev
Build date: 17:06:18 Mar 15 2017 (1068)
Build from: https://github.com/dreamstalker/rehlds/commit/f943efb
```

amxx plugins
```
Currently loaded plugins:
       name                    version     author            file             status   
 [  1] Report Abuse            0.0.1       MaZDuR            report_abuse.am  running  
 [  2] AMXX Gag                1.5.0       xPaw & Exolent    amx_gag.amxx     running  
 [  3] uneven teams warning    1.2         pizzahut          utw.amxx         running  
 [  4] Advanced Bans           0.9.8       Master            advanced_bans_t  running  
 [  5] Evil Network Plugin     0.2.8       Master            evil_server.amx  debug    
 [  6] Skynet                  0.6.0       Master            skynet.amxx      running  
 [  7] amx_who                 1.0         Yami Kaitou       amxwho.amxx      running  
 [  8] unknown                 unknown     unknown           command_logger.  bad load 
 [  9] Admin Base (SQL)        1.8.3-dev+  AMXX Dev Team     admin_sql.amxx   running  
 [ 10] Admin Commands          1.8.3-dev+  AMXX Dev Team     admincmd.amxx    running  
 [ 11] Admin Help              1.8.3-dev+  AMXX Dev Team     adminhelp.amxx   running  
 [ 12] Slots Reservation       1.8.3-dev+  AMXX Dev Team     adminslots.amxx  running  
 [ 13] Multi-Lingual System    1.8.3-dev+  AMXX Dev Team     multilingual.am  running  
 [ 14] Menus Front-End         1.8.3-dev+  AMXX Dev Team     menufront.amxx   running  

[...]

PR #687, opened 2019-03-22
fix buffer overflow

namazso:
from the new cs update

### DIFF
--- a/rehlds/engine/sv_main.cpp
+++ b/rehlds/engine/sv_main.cpp
@@ -5833,7 +5833,7 @@ void EXT_FUNC SV_ActivateServer_internal(int runPhysics)
 	if (mapchangecfgfile.string && *mapchangecfgfile.string)
 	{
 		AlertMessage(at_console, "Executing map change config file\n");
-		Q_sprintf(szCommand, "exec %s\n", mapchangecfgfile.string);
+		Q_snprintf(szCommand, sizeof(szCommand), "exec %s\n", mapchangecfgfile.string);
 		Cbuf_AddText(szCommand);
 	}
 


DarthMan:
Was the TFC bug fixed as well?

namazso:
no, this is a single line fix that i stumbled upon when diffing old and new client versions to see how the recent malware worked. If whatever TFC bug you're talking about still exists, you can diff tfc versions and see what was changed

DarthMan:
https://github.com/dreamstalker/rehlds/issues/685

Read here. First my guess was that a plug-in caused it, but even without amxx and metamod enabled this problem still persists. With HLDS this doesn't happen and other ppl had the same issue. So it clearely is a problem with ReHLDS which must be fixed because i'd love to use Re for my TFC server as well.

MaZDuR:
#423 
#606 

I believe the issue is because rehlds was built against a broken incompatible dll that no server owners used, there is a new dll/so file being released tomorrow, I am unsure how rehlds works but this is indeed a problem that nobody has yet addressed, the game simply will not work with rehlds.

DarthMan:
> 
> 
> #423
> #606
> 
> I believe the issue is because rehlds was built against a broken incompatible dll that no server owners used, there is a new dll/so file being released tomorrow, I am unsure how rehlds works but this is indeed a problem that nobody has yet addressed, the game simply will not work with rehlds.

Let's just hope that it will be fixed. When Valve will release the bugfixed tfc.dll / tc.so (still in beta) I'll make the switch but until then I'll be using the old dll because I'm using Orpheu in some of my plug-ins and I'll have to make new signatures (due dll's being compiled with a newer version of Visual Studio after the SteamPipe update) which will take some time.

Issue #680, opened 2019-01-20
Graceful shutdown on sigterm


Maxpain177:
Hi. I think ReHLDS should implement graceful shutdown in `hlds_linux` launcher when receives `SIGINT` / `SIGTERM` signals.
In SRCDS it's working normally. When a server receives this signal, all players are kicking with the appropriate message.

It's very important to implement graceful shutdown in environments like Docker with Kubernetes, etc.
Sending `exit` command to the stdin it's the ugly workaround.

Maxpain177:
Any updates?

Maxpain177:
It's so important for our cloud of game servers

justgo97:
did you try this ?

http://teamcity.rehlds.org/repository/download/Rehlds_Publish/20746:id/rehlds-dist-3.7.0.697-dev.zip

Maxpain177:
Any news?

PR #700, opened 2019-06-21
Implement sigterm & sigkill handler

fant1kua:
Trying to implement [#680](https://github.com/dreamstalker/rehlds/issues/680) sigterm and sigkill handler with allow graceful shutdown. It will be helpful for docker containers and other stuff witch send signals for closing server.

Credits: [In-line](https://github.com/In-line) who help me a lot with it

### DIFF
--- a/rehlds/dedicated/src/precompiled.h
+++ b/rehlds/dedicated/src/precompiled.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <atomic>
+
 #include "basetypes.h"
 #include "FileSystem.h"
 #include "strtools.h"
--- a/rehlds/dedicated/src/sys_ded.cpp
+++ b/rehlds/dedicated/src/sys_ded.cpp
@@ -30,7 +30,7 @@
 
 bool g_bVGui = false;
 char *gpszCvars = nullptr;
-bool g_bAppHasBeenTerminated = false;
+std::atomic<bool> g_bAppHasBeenTerminated(false);
 
 CSysModule *g_pEngineModule = nullptr;
 IDedicatedServerAPI *engineAPI = nullptr;
@@ -160,7 +160,7 @@ int RunEngine()
 
 		Sys_PrepareConsoleInput();
 
-		if (g_bAppHasBeenTerminated)
+		if (g_bAppHasBeenTerminated.load(std::memory_order_acquire))
 			break;
 
 #ifdef VGUI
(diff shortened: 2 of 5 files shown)


rehlds/dedicated/src/sys_linux.cpp
```diff
@@ -327,7 +339,6 @@ int main(int argc, char *argv[])
 {
 	Q_snprintf(g_szEXEName, sizeof(g_szEXEName), "%s", argv[0]);
 	char* cmdline = BuildCmdLine(argc, argv);
-	StartServer(cmdline);
-
-	return 0;
+	auto ret = StartServer(cmdline);
+	return ret == LAUNCHER_OK ? 0 : 1;
```
In-line: Maybe REHLDS_FIXES macro? Can you remove `ret` variable?

rehlds/dedicated/src/sys_linux.cpp
```diff
@@ -314,8 +314,20 @@ char* BuildCmdLine(int argc, char **argv)
...
+{
+	if (signal == SIGINT || signal == SIGTERM)
+	{
+		g_bAppHasBeenTerminated.store(true, std::memory_order_release);
+	}    
+}
+
 bool Sys_SetupConsole()
 {
+	struct sigaction action;
```
In-line: Add one more space

rehlds/dedicated/src/precompiled.h
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <atomic>
```
In-line: Did you verify that this doesn't pull dependency on newer `libstdc++`?

rehlds/dedicated/src/sys_ded.cpp
```diff
@@ -30,7 +30,7 @@
 
 bool g_bVGui = false;
 char *gpszCvars = nullptr;
-bool g_bAppHasBeenTerminated = false;
+std::atomic<bool> g_bAppHasBeenTerminated(false);
```
In-line: I think `atomic<..>` is necessary anyway. Because in linux there is no guarantee which thread will receive the signal.

rehlds/dedicated/src/sys_ded.cpp
```diff
@@ -193,14 +193,14 @@ int RunEngine()
 	VGUIFinishedConfig();
 
 	if (iret == DLL_CLOSE)
-		g_bAppHasBeenTerminated = true;
+		g_bAppHasBeenTerminated.store(true, std::memory_order_release);
```
In-line: Maybe we can relax atomic more, but simple `acquire`/`release` seems to be okay

rehlds/dedicated/src/sys_linux.cpp
fant1kua: Done

rehlds/dedicated/src/sys_linux.cpp
fant1kua: Done

rehlds/dedicated/src/precompiled.h
fant1kua: I tried to start server and it works fine

Maxpain177:
Any news?

Issue #699, opened 2019-06-18
ReHLDS Crash


DarthMan:
Hello there. Everytime I connect to a non-CS server that was hosted using the latest ReHLDS dev builds, the server seems to crash. It works fine until a player connects, then it crashes with no errors. The build numbers that are causing these crashes are ending in 77 and 78.

Infamous2017:
you should add more infos.  meta list , meta version.   without those things we cant help you.

DarthMan:
The server crashes even without metamod and AMXX running, so giving this info is useless. So if my HL1 server has as parameter in liblist.gam dlls/hl.dll it still crashes right when a player joins. That clearely means it has nothing to do with AMXX plug-ins or metamod version.

DarthMan:
I'm using Windows Server 2008, if that's of any help.

Infamous2017:
Which Reunion Version u are using?  Or dproto ?   I remember long time ago i had this problem with an HLDS Server + Old Dproto.

DarthMan:
As I said, I'm not using metamod. So that means that I'm also not using dproto / ReUnion. And yes, all of my servers are Steam-only. This doesn't happen with official ReHLDS, just with latest unnofficial branches.

eat1k:
> As I said, I'm not using metamod. So that means that I'm also not using dproto / ReUnion. And yes, all of my servers are Steam-only. This doesn't happen with official ReHLDS, just with latest unnofficial branches.

So it’s not a problem of ReHLDS if you’re not using the official latest ReHLDS version.

DarthMan:
> 
> 
> > As I said, I'm not using metamod. So that means that I'm also not using dproto / ReUnion. And yes, all of my servers are Steam-only. This doesn't happen with official ReHLDS, just with latest unnofficial branches.
> 
> So it’s not a problem of ReHLDS if you’re not using the official latest ReHLDS version.

Fixed with latest dev build .700

eat1k:
Just to clarify, the latest successful build is 3.6.0.672. It's the version that should be used to have no issues.

Issue #698, opened 2019-06-08
Windows 2008 cause crashes


BlackShadow306:
This issue was happening in original HLDS as well. reHLDS crashing randomly without any log in Windows 2008.

voed:
Provide crash dumps.

BlackShadow306:
Sadly as i stated this crash doesn't generate any log regarding the issue.

krangm:
@BlackShadow306 use dumper
http://www.microsoft.com/en-us/download/details.aspx?id=4060

BlackShadow306:
Alright, i'll try and let's see what can i get from it. Thanks!

BlackShadow306:
Can't install dumper on Win 2008.

krangm:
try WER
https://docs.microsoft.com/en-us/windows/desktop/wer/collecting-user-mode-dumps

BlackShadow306:
Well i got some dumps, hopefully it will be usefull for you guys:

https://drive.google.com/open?id=1VaY3QsKC21kSkZIppqhLBX0A61D9l0j_

Issue #523, opened 2017-09-14
Crashes: After playing a few minutes


Zetz512:
Hi!!
I have a issue that crashes after about 30 min, more or less of playing...
with this error:

IMG: https://gyazo.com/7b71274fe440ccae93b3ef160ffcc10c

>  
Meta list:     description      stat pend  file              vers      src   load  unlod
 [ 1] AMX Mod X        RUN   -    amxmodx_mm.dll    v1.8.3-d  ini   Start ANY  
 [ 2] Reunion          RUN   -    reunion_mm.dll    v0.1.75   ini   Start Never
 [ 3] VoiceTranscoder  RUN   -    voicetranscoder.  v2017RC3  ini   ANY   ANY  
 [ 4] ReAuthCheck      RUN   -    reauthcheck_mm.d  v0.1.6    ini   Start Never
 [ 5] Rechecker        RUN   -    rechecker_mm.dll  v2.3      ini   Chlvl ANY  
 [ 6] ReAPI            RUN   -    reapi_amxx.dll    v5.1.0.1  pl1   ANY   Never
 [ 7] ReAimDetector    RUN   -    reaimdetector_am  v0.1.7    pl1   ANY   Never
 [ 8] Ham Sandwich     RUN   -    hamsandwich_amxx  v1.8.3-d  pl1   ANY   ANY  
 [ 9] CSX              RUN   -    csx_amxx.dll      v1.8.3-d  pl1   ANY   ANY  
 [10] FakeMeta         RUN   -    fakemeta_amxx.dl  v1.8.3-d  pl1   ANY   ANY  
 [11] hackdetector     RUN   -    hackdetector_amx  v0.15.32  pl1   ANY   ANY  
 [12] CStrike          RUN   -    cstrike_amxx.dll  v1.8.3-d  pl1   ANY   ANY  
 [13] Engine           RUN   -    engine_amxx.dll   v1.8.3-d  pl1   ANY   ANY  
13 plugins, 13 running

Status:
version :  48/1.1.2.7/Stdio 1250 insecure (secure mode enabled, connected to Steam3)(u) (10)
tcp/ip  :  25.8.245.15:27015
map     :  de_dust2 at: 0 x, 0 y, 0 z
players :  6 active (32 max)

game version
ReGameDLL version: 5.3.0.221-dev
Build date: 17:13:45 Aug 19 2017
Build from: https://github.com/s1lentq/ReGameDLL_CS/commit/462dc99

meta version
[...]

PR #703, opened 2019-07-01
Catch crashes and write minidumps on windows

dimhotepus:
Kinda hard to debug issues without crash dumps on hands.  Add crash dumps writing support on Windows. Makes resolving issues like https://github.com/dreamstalker/rehlds/issues/523, https://github.com/dreamstalker/rehlds/issues/698, https://github.com/dreamstalker/rehlds/issues/699 easier.

Crash dump is written to file in the same directory as crashed executable.  Crash dump file name format: `<exe name>_<app version>_crash_<date YYYYMMDD>_<time HHMMSS>_<unique counter>.mdmp`
for example, `hlds_3.6.0.675-dev+m_crash_20190702_005019_1.mdmp`

Command line option `-nominidumps` disables mini dump writing facility (same behavior as in Source games). Also, mini dumps are not written when debugger is attached.

Uses `DbgHelp.dll` to write minidumps, should be present on all Windows versions, but **REQUIRES checking** on Windows XP+. Tested on Windows 10 18362.

Tries to write minidump with data segments, handles, threads and indirectly referenced memory first.
As a fallback, writes minidump with data segments only.

### DIFF
--- a/rehlds/common/minidump.cpp
+++ b/rehlds/common/minidump.cpp
@@ -4,10 +4,49 @@
 #if defined(HLTV_FIXES) || defined(LAUNCHER_FIXES)
 #ifdef _WIN32
 
+#include <DbgHelp.h>
+
 // Gets last error.
 static inline HRESULT GetLastHresult()
 {
 	return HRESULT_FROM_WIN32(GetLastError());
+}
+
+#define WIDE_TEXT_HELPER_(quote) L##quote
+#define WIDE_TEXT(quote) WIDE_TEXT_HELPER_(quote)
+
+// Gets application version which is safe for use in minidump file name.  Can be entirely constexpr with C++17 and std::array.
+template<size_t appVersionSize>
+static wchar_t* GetAppVersionForMiniDumpName(wchar_t (&appVersion)[appVersionSize]) {
+	constexpr wchar_t rawAppVersion[]{WIDE_TEXT(APP_VERSION)};
+
+	static_assert(appVersionSize >= ARRAYSIZE(rawAppVersion), "App version buffer size should be enough to store app version.");
+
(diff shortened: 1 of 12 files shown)


dimhotepus:
@WPMGPRoSToTeMa, @theAsmodai Can you, please, suggest how to fix missing symbols during link step?

Locally all builds and works ok (but i've converted solution to vs 2019 format + c++14).

rehlds/common/minidump.cpp
```diff
@@ -0,0 +1,135 @@
...
+			local_time_now->tm_mday,        /* day of month (1 - 31) */
+			local_time_now->tm_hour,        /* hour (0 - 23) */
+			local_time_now->tm_min,         /* minutes (0 - 59) */
+			local_time_now->tm_sec,         /* seconds (0 - 59) */
+			++minidumps_written_count);     // ensures the filename is unique
+
+		minidump_file = CreateFileW(file_name, GENERIC_WRITE, FILE_SHARE_WRITE,
+									nullptr, CREATE_ALWAYS,
+									FILE_ATTRIBUTE_NORMAL, nullptr);
+		error_code = !!minidump_file ? S_OK : GetLastHresult();
```
tsarevichdm: Please, fix check, `minidump_file` will be `INVALID_HANDLE_VALUE` when `CreateFileW` failed.

See https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew#return-value

dimhotepus: Done, thank you.

In-line:
Seems to be okay at the first glance, build should be fixed and I think this can be merged hopefully. I don't run windows, to manually test.

dimhotepus:
@In-line Done, thanks!

dimhotepus:
@s1lentq Thank you for fixing problems. Commit hash is good idea, too.

PR #708, opened 2019-08-04
Small update sv_use_entity_file

Karaulov:
Now if sv_use_entity_file set to 1, don't create new .ent files, use only exist.


Karaulov:
Possible need fix text in readme, i don't know english, I use google translate :)

PR #710, opened 2019-08-04
Adding cvar sv_rcon_allowexternal to allow or disallow external usage of Rcon

GiovaniFerraroTrivelli:
Some panels, like OpenGamePanel, use a option to send commands through Rcon. In my case, for security, I disabled Rcon (rcon_password "") but I need an option to send commands without entering to the server.

I added a cvar, sv_rcon_allowexternal (0/1) that allows or disallows the usage of Rcon from externals IPs. Also, it logs when cvar is 0 and someone tries to use it.

This is more simple than #587, which is a very good idea. In the future it could be improved to match this suggestion.


RauliTop:
Don't forget to add on the README.md

GiovaniFerraroTrivelli:
Thanks!

LevShisterov:
Hi! Thanks for changes! Please split them in different PRs. Particulary, this one is supposed to contain only `sv_rcon_allowexternal` related code.

GiovaniFerraroTrivelli:
Sorry, my bad! I forgot I did this PR. I'm going to open a new pull request correctly.

PR #714, opened 2019-08-17
Another entities improvement!

Karaulov:
Added sv_use_global_entity_file for using your custom entities in all maps! (for example for AMXX plugins)

Tested, it works.

Possible need fixes readme or code.


RauliTop:
An example where this will be used?

Karaulov:
For example you can add vehicle (https://dev-cs.ru/threads/4429/post-64042) in global entities, and using amxx script spawn it on any map.

Karaulov:
or how to use ED_LoadFromFile function from AMXX ?

PR #719, opened 2019-09-04
Cbuf_Execute: Add checks commented out for multi-lines.

s1lentq:
## Description

Cbuf_Execute doesn't ignore commented out string in case when string contains separator character `;` inside commented out string.

## Steps to reproduce

1. Config file `test.cfg` contains 
>// ; echo "test: commented out string"
2. Execute `exec test.cfg`
3. The text will printed in the console.

### DIFF
--- a/rehlds/engine/cmd.cpp
+++ b/rehlds/engine/cmd.cpp
@@ -164,8 +164,9 @@ void Cbuf_Execute(void)
 {
 	int i;
 	char *text;
-	char line[MAX_CMD_LINE];
+	char line[MAX_CMD_LINE] = {0};
 	int quotes;
+	bool commented;
 
 	while (cmd_text.cursize)
 	{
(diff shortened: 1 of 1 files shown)


rehlds/engine/cmd.cpp
```diff
@@ -164,21 +164,31 @@ void Cbuf_Execute(void)
...
 	{
 		// find a \n or ; line break
 		text = (char *)cmd_text.data;
 
 		quotes = 0;
+		commented = false;
+
+#ifdef REHLDS_FIXES
+		if (cmd_text.cursize > 1 && text[0] == '/' && text[1] == '/')
+			commented = true;
```
afwn90cj93201nixr2e1re: Maybe we should also add ';' as comment line?

voed: > Maybe we should also add ';' as comment line?

`;` is a command separator, it cannot be comment character

afwn90cj93201nixr2e1re: Не, ну какая команда начинается с `;` в начале строки?

PR #720, opened 2019-09-04
Fix _vsnprintf on windows

FFx0q:
warning C4996: _vsnprintf': This function or variable may be unsafe. Consider using _vsnprintf_s instead.

### DIFF
--- a/rehlds/public/tier0/dbg.h
+++ b/rehlds/public/tier0/dbg.h
@@ -422,7 +422,7 @@ public:
 		va_list arg_ptr;
 
 		va_start(arg_ptr, pszFormat);
-#ifdef _WIN32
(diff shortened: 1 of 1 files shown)


s1lentq:
You should be use macro _CRT_SECURE_NO_WARNINGS too

rehlds/public/tier0/dbg.h
```diff
@@ -422,7 +422,11 @@ class CDbgFmtMsg
 		va_list arg_ptr;
 
 		va_start(arg_ptr, pszFormat);
+#ifdef _WIN32
```
s1lentq: You should use macro `!defined(_CRT_SECURE_NO_WARNINGS)` too

FFx0q: Okey, thanks

PR #737, opened 2019-12-01
Implement `svc_exec` support in the engine and HLTV

### LINKED ISSUE #734: [TFC] SVC_EXEC - Message Unregistered
A new SVC_EXEC message was added to TFC to exec class specific cfg files on clients in a recent update. But since ReHLDS is based on an earlier version, class cfgs no longer auto exec at spawn. Without this, clients need to manually exec a cfg every time they switch class, which a pain in the ass.

Other than that, everything else seemed to work well and I would probably use ReHLDS if this were to be fixed.

WPMGPRoSToTeMa:
Resolves #734.

### DIFF
--- a/rehlds/HLTV/Core/src/Server.cpp
+++ b/rehlds/HLTV/Core/src/Server.cpp
@@ -378,7 +378,12 @@ void Server::ProcessMessage(unsigned int seqNr)
 			break;
 		}
 
+// With `HLTV_FIXES` enabled meaning of `svc_startofusermessages` changed a bit: now it is an id of the first user message
+#ifdef HLTV_FIXES
+		if (cmd >= svc_startofusermessages)
+#else // HLTV_FIXES
 		if (cmd > svc_startofusermessages)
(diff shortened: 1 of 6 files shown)


LevShisterov:
Probably [net_internal.h](https://github.com/dreamstalker/rehlds/blob/65c6ce593b5eabf13e92b03352e4b429d0d797b0/rehlds/HLTV/common/net_internal.h#L162) should be changed too.

WPMGPRoSToTeMa:
> 
> 
> Probably [net_internal.h](https://github.com/dreamstalker/rehlds/blob/65c6ce593b5eabf13e92b03352e4b429d0d797b0/rehlds/HLTV/common/net_internal.h#L162) should be changed too.

Changed it. @LevShisterov, @s1lentq please take a look at my latest changes.

s1lentq:
it seems to be fine

WPMGPRoSToTeMa:
@s1lentq done, note that `64` is an id of the first user message, not the last svc_*.

Issue #744, opened 2020-01-25
[HLTV][FRequest] New command: chatdelay


wopox1337:
Need to control delay between HLTV client chat messages.
i suggest add new cmd for configure that: `chatdelay <0 ... 999>`

https://github.com/dreamstalker/rehlds/blob/65c6ce593b5eabf13e92b03352e4b429d0d797b0/rehlds/HLTV/Proxy/src/ProxyClient.cpp#L129

PR #777, opened 2020-08-19
Add: new chatdelay command.

wopox1337:
Close https://github.com/dreamstalker/rehlds/issues/744

### DIFF
--- a/rehlds/HLTV/Proxy/src/Proxy.cpp
+++ b/rehlds/HLTV/Proxy/src/Proxy.cpp
@@ -2777,7 +2777,7 @@ const char *Proxy::GetDescription()
 	return "Private Server";
 }
 
-void Proxy::CMD_ChatDelay(char* cmdLine)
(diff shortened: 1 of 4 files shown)


rehlds/HLTV/Proxy/src/Proxy.cpp
```diff
@@ -188,6 +189,7 @@ bool Proxy::Init(IBaseSystem *system, int serial, char *name)
 	m_MaxClients = 128;
 	m_MaxQueries = 100;
 	m_Region = 255;
+	m_ChatDelay = 6;
```
theAsmodai, line 192: default value should be 0

afwn90cj93201nixr2e1re: .

theAsmodai: ok

PR #806, opened 2021-02-08
static_map.h: fix lowercase convert

2020karaulov2020:
**REHLDS_OPT_PEDANTIC** static_map.h fix

`if (true) {	cc |= 0x20; } `
to 
`if ( cc >= 'A' && cc <= 'Z') { cc |= 0x20;	}`

### DIFF
--- a/rehlds/public/rehlds/static_map.h
+++ b/rehlds/public/rehlds/static_map.h
@@ -229,7 +229,7 @@ protected:
 		if (cpuinfo.sse4_2) {
 			while (*pcc) {
 				char cc = *(pcc++);
-				if (cc >= 'A' || cc <= 'Z') {
+				if (cc >= 'A' && cc <= 'Z') {
 					cc |= 0x20;
 				}
 				cksum = crc32c_t8_sse(cksum, cc);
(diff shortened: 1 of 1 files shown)


2020karaulov2020:
NEED ?

RauliTop:
seems to be a typo

rehlds/public/rehlds/static_map.h
```diff
@@ -229,8 +229,8 @@ class CICaseStringKeyStaticMap : public CStaticMap<const char*, T_VAL, ASSOC_2N,
 		if (cpuinfo.sse4_2) {
 			while (*pcc) {
 				char cc = *(pcc++);
-				if (cc >= 'A' || cc <= 'Z') {
-					cc |= 0x20;
+				if (cc >= 'A' && cc <= 'Z') {
+					cc += ('a' - 'A');
```
WPMGPRoSToTeMa, line 233: IMO that is more readable:
```suggestion
					cc = (cc - 'A') + 'a';
```

wopox1337: I would like to use Q_strlwr but....

wopox1337 (approved):

rehlds/public/rehlds/static_map.h
wopox1337: I would like to use tolower() but....

Thanks. Done b7809bd

Issue #841, opened 2021-06-23
VirusTotal detection


wopox1337:
from: https://dev-cs.ru/threads/131/page-22#post-108131

detection: https://www.virustotal.com/gui/file/e754f765ad6d253a6551d700af5f0cacdcd31933731d7b762a7434a03a006fff/detection

![image](https://user-images.githubusercontent.com/18553678/123081880-19682200-d427-11eb-8f03-526327bfecf1.png)

Problems in:
-  [filesystem_stdio.dll](https://www.virustotal.com/gui/file/6847262be2e67da2af5b4cffd77dc0a15a8a5e08d63a1b554a5bc75e20cc9c74/detection)
- [hltv.exe](https://www.virustotal.com/gui/file/6847262be2e67da2af5b4cffd77dc0a15a8a5e08d63a1b554a5bc75e20cc9c74/detection)

hajimura:
Should be closed as fixed

jeefo:
As for easy fix, you can switch to v141_xp toolset, and inject version-info for each binary in windows build.

Result: https://www.virustotal.com/gui/file/43bd19869fc264a177bfc7b6c3c54fd12da9c505da944538bdac0d47e9fa3de7/details

As for hltv.exe (and occasionally hlds.exe), these files should be signed with valid certificate to get more trust from AV vendors.

stambeto2006:
@wopox1337 When will you fix the problem so that we can use the latest versions for Windows as well, not just Linux?

stambeto2006:
@s1lentq @wopox1337 When will you fix the problem so that we can use the latest versions for Windows as well, not just Linux?

TheStormN:
@wopox1337 Can you just separate the packages?

For example have: `rehlds-bin-win32-VERSION_NUM.zip` and `rehlds-bin-linux-VERSION_NUM.zip`?

This way the Windows build will still be easily downloadable and the Linux version will not be flagged.

PR #842, opened 2021-06-23
change destinition folder for linux build

wopox1337:
Temporary solution for VirusTotal issues.

related #841


torriz:
Are you planning on adding back windows builds? @wopox1337 
All the builds after this version is missing windows bin

PR #894, opened 2022-01-31
Fix usage of sv_rehlds_maxclients_from_single_ip

ShadowsAdi:


rehlds/engine/sv_main.cpp
```diff
@@ -1901,7 +1901,7 @@ int SV_CheckIPConnectionReuse(netadr_t *adr)
 	client_t *cl = g_psvs.clients;
 	for (int i = 0; i < g_psvs.maxclients; i++, cl++)
 	{
-		if (cl->connected && !cl->fully_connected && NET_CompareBaseAdr(cl->netchan.remote_address, *adr))
```
ShadowsAdi, line 1904: Connecting clients will be also checked: https://github.com/dreamstalker/rehlds/blob/master/rehlds/engine/sv_main.cpp#L2491

RauliTop:
> 

a pull request to change description so?
Don't change the behavior

Issue #888, opened 2022-01-12
Avatar not showing after hooking SV_ConnectClient function


ShadowsAdi:
Hello,
I am using [this plugin](https://github.com/ShadowsAdi/AdvancedSlotReservation/blob/master/scripting/advanced_slot_reservation.sma) which is hooking [SV_ConnectClient](https://github.com/dreamstalker/rehlds/blob/master/rehlds/engine/sv_main.cpp#L2261-L2264) in pre state. If this condition [check](https://github.com/ShadowsAdi/AdvancedSlotReservation/blob/master/scripting/advanced_slot_reservation.sma#L279-L280) is passed, a connected players is kicked out from the server and when the new one is joining it does not show his steam avatar. Otherwise, if the check is not passed, it will let the player connect ( because server is not full ) and his steam avatar will show.

Ant1Lamer:
Avatar and show STEAM_ID_LAN

PR #896, opened 2022-02-05
Netchan_FlushIncoming: Don't clear global buffer of incoming network messages

s1lentq:
Issue related to #888 is that, what we can't call SV_DropClient for another client inside callback while handling incoming network messages,
because call chains are `SV_DropClient -> Netchan_Clear -> Netchan_ClearFragments -> Netchan_FlushIncoming -> msg_readcount` cause clearing `msg_readcount` and `net_message` for all.

```
- SV_ConnectClient PRE  (id 1) net_message[msg_readcount] - there is something to read here
- SV_ConnectClient 3rd party   call SV_DropClient (id 2) -> msg_readcount = 0
- SV_ConnectClient POST (id 1) net_message[msg_readcount] - we have nothing left to read for this client, steam ticket is empty
```

Every incoming message initializes these global variables, so do we really need to clean them up while flushing?

### DIFF
--- a/rehlds/engine/net_chan.cpp
+++ b/rehlds/engine/net_chan.cpp
@@ -1389,10 +1389,13 @@ void Netchan_FlushIncoming(netchan_t *chan, int stream)
 {
 	fragbuf_t *p, *n;
 
-#ifndef REHLDS_FIXES
-	SZ_Clear(&net_message);
-	msg_readcount = 0;
+#ifdef REHLDS_FIXES
+	if (chan->player_slot == host_client - g_psvs.clients)
 #endif
+	{
+		SZ_Clear(&net_message);
+		msg_readcount = 0;
+	}
 
 	p = chan->incomingbufs[stream];
 	while (p)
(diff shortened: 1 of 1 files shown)


rehlds/engine/net_chan.cpp
```diff
@@ -1389,8 +1389,10 @@ void Netchan_FlushIncoming(netchan_t *chan, int stream)
 {
 	fragbuf_t *p, *n;
 
+#ifndef REHLDS_FIXES
 	SZ_Clear(&net_message);
 	msg_readcount = 0;
+#endif
```
WPMGPRoSToTeMa, line 1395: I think this is needed to prevent reading messages from the dropped client (i.e. a client was dropped when reading in the middle of the buffer). It would be better to check if we're dropping the current client:
```suggestion
#ifdef REHLDS_FIXES
	if (chan->player_slot == host_client - g_psvs.clients)
#endif
	{
		SZ_Clear(&net_message);
		msg_readcount = 0;
	}
```

WPMGPRoSToTeMa (changes requested):

Issue #855, opened 2021-08-07
Crash with SZ_GetSpace overflow on (Nick)


RauliTop:
Recently I added some new features to my server.

I don't know where is the problem, but every 2 days server is crashing with console overflowed with this message: SZ_GetSpace overflow on (Nick)
All the time on the same client, 3512 lines of messages.

Yes, it's an overflow on the client.
But client isn't dropped and result an overflow on the server (crash).

ReHLDS should add a protection.
The client may be kicked when 'X' amount of prints done, that's to prevent server crash.

How to kick client? Is it only possible inside ReHLDS?
That's where protection must be applied: https://github.com/dreamstalker/rehlds/blob/master/rehlds/engine/common.cpp#L1241

Maybe change?
`Con_Printf("%s: overflow on %s\n", __func__, buffername);`
to
`Con_NetPrintf("%s: overflow on %s\n", __func__, buffername);`
Like in this commit: https://github.com/dreamstalker/rehlds/commit/98db4672cf90f82052888ae5a5c0fb71a7298dce

RauliTop:
How to reduce and solve this problem: https://forums.alliedmods.net/showthread.php?t=334122

RauliTop:
Well, adding a plugin with Con_Printf hook for block the console print and finally kick the player when too much times printing, seems to be working without crash.

The problem was when the overflow is caused by: netchan -> message.
We can't know who is causing this.

So, the only way will be testing the change to Con_NetPrintf.

**Anybody to make the pull request?**

Issue #798, opened 2021-01-14
SZ_GetSpace: overflow on 


stambeto2006:
Constantly spam from the floods of these things as a message. 
- Zombie Plague Mod

**SZ_GetSpace: overflow on Player
SZ_GetSpace: overflow on Player
SZ_GetSpace: overflow on Player
SZ_GetSpace: overflow on Player
SZ_GetSpace: overflow on Player
WARNING: datagram overflowed for Player**

2020karaulov2020:
Same problem, need fully fix this issue!

stambeto2006:
Is there any development on this issue?

wopox1337:
just disable metamod and this issue will fixed.

or try to disable one-by-one amxx plugins. This problem isn't from ReHLDS engine


Moderator, please close this topic.

WPMGPRoSToTeMa:
This usually appears when a player has temporarily lost the connection, not when the server sends a lot of data. Time elapsed from the last packet from the client can be added to this warning message to indicate this situation.

RauliTop:
> This usually appears when a player has temporarily lost the connection, not when the server sends a lot of data. Time elapsed from the last packet from the client can be added to this warning message to indicate this situation.

dproto already has it.

Will be nice to have on rehlds too.

PR #907, opened 2022-04-08
TESTING: fix crash due to SZ_GetSpace console overflow

RauliTop:
More info here: https://github.com/dreamstalker/rehlds/issues/855

Also, review this: https://github.com/dreamstalker/rehlds/issues/798

### DIFF
--- a/rehlds/engine/common.cpp
+++ b/rehlds/engine/common.cpp
@@ -1238,7 +1238,7 @@ void *EXT_FUNC SZ_GetSpace(sizebuf_t *buf, int length)
 		}
 #endif // REHLDS_FIXES
 
-		Con_Printf("%s: overflow on %s\n", __func__, buffername);
+		Con_NetPrintf("%s: overflow on %s\n", __func__, buffername);
 
 		SZ_Clear(buf);
 		buf->flags |= SIZEBUF_OVERFLOWED;


Kasou123:
help SZ_GetSpace: overflow on gs-m <- russian
WARNING: datagram overflowed for gs-m <- russian

stambeto2006:
Is there any solution to this problem with this message, and how is the problem fixed?

overflow on netchan->message

stambeto2006:
@RauliTop Is this project expected to be implemented and fixed officially soon?

RauliTop:
> @RauliTop Is this project expected to be implemented and fixed officially soon?

I really didn't test it.
I'm currently doing changes on plugins, without successful results.

You can try my request downloading the binaries on Actions tab inside this repository.
Pull request has been updated now with lastest rehlds official changes.

Alguiencomoyo:
> More info here: #855
> 
> Also, review this: #798

my servers go down with that same message I discovered that it is an attacker
even then I enter the servers to ask for money in exchange so as not to continue attacking
I have everything updated rehlds, amx, etc
and even so your proposal can solve this too?
I attach some videos of how he does it

https://www.youtube.com/watch?v=tCQ0tFdUTho

https://www.youtube.com/watch?v=Ex4n9KfUiCQ

https://www.youtube.com/watch?v=WMDZvrewaNI

pepepepito0147:
also affects? [rehlds/engine/cmd.cpp#L122](https://github.com/dreamstalker/rehlds/blob/f955b07b69a10db72290b2b35e3579151e9dc810/rehlds/engine/cmd.cpp#L122)

RauliTop:
> also affects? [rehlds/engine/cmd.cpp#L122](https://github.com/dreamstalker/rehlds/blob/f955b07b69a10db72290b2b35e3579151e9dc810/rehlds/engine/cmd.cpp#L122)

No, because after overflow it appears: 'on' %s
It's on the line I marked.

mlibre2:
Well this problem is quite common, apparently one of the causes is the loss of the player's connection to the server. `sv_timeout` is a quick fix for this, but I wanted to find another one, since this cvar doesn't work quite right, so I wrote this [plugin](https://github.com/mlibre2/pawnamxx/blob/main/scripting/sv_timeout.sma) to check the communication with the player. This method has worked well for me so far, although I don't like to call a function frequently, maybe there is another more optimal way to do it, but at least it doesn't print the typical "overflow/datagram" error message to the console, which that prevents it from happening. 

<sub>So far it has been ruled out that the false positives have been kicked without reason!</sub>

anzz1:
This is wrong, it does not match original code.

There is no `Con_NetPrintf` in `SZ_GetSpace` in original HLDS. If this is a ReHLDS fix, it should be wrapped in `#ifdef REHLDS_FIXES` appropriately. 

Also I don't think Con_Printf has anything to do with the error.

PR #937, opened 2022-09-05
Added new cvars "sv_invalid_length" and "sv_msg_badread"

mlibre2:
Fix bug to fake players -> **sv_invalid_length**

Disable message flooding in console -> **sv_msg_badread**


RauliTop:
Add reauthcheck to your server, and forget about this issue

pepepepito0147:
> 

Yes indeed it is, but the idea is to _allow_ them to connect, _reauthcheck_ already prevents it since before this update, so added cvar to enable or disable this for whoever wants or doesn't allow these types of fake players to access to the server.

PR #942, opened 2022-09-22
Minor changes and code refactor

GermanAizek:
@s1lentq hello, review PR changes and feedback me.

### DIFF
--- a/rehlds/HLTV/Core/src/BSPModel.h
+++ b/rehlds/HLTV/Core/src/BSPModel.h
@@ -45,7 +45,7 @@ class IBaseSystem;
 class BSPModel: public IBSPModel {
 public:
 	BSPModel() {}
-	virtual ~BSPModel() {}
+	virtual ~BSPModel() { delete m_wadpath; }
 
 	EXT_FUNC void Init(IBaseSystem *system);
 	EXT_FUNC void Clear();


fred0r:
been using this pr for some weeks now on linux with regamedll_cs and the hitboxes 'feel' much more 'accurate/natural' than before.
the downside is that the hltv  often gets disconnected with `Dropped from x.x.x.x:27015 (Reliable channel overflowed).`
in my setup the server records every map played.

justgo97:
> been using this pr for some weeks now on linux with regamedll_cs and the hitboxes 'feel' much more 'accurate/natural' than before. the downside is that the hltv often gets disconnected with `Dropped from x.x.x.x:27015 (Reliable channel overflowed).` in my setup the server records every map played.

The PR doesn't include any changes to how hitboxes function.

GermanAizek:
@fred0r @justgo97 I had already forgotten about this PR, it turns out that someone else answered me. HLTV breakdown is possible due to `delete` operator or additional checks, I can divide this PR into two parts, where refactoring and where fixes.

GermanAizek:
@justgo97 rehlds project is still active, are you doing it? In September, I created a PR, but no one answered it.

justgo97:
> @justgo97 rehlds project is still active, are you doing it? In September, I created a PR, but no one answered it.

I'm not the project manager, I'm a normal user like you, I guess you can try to divide the PR to address one issue at a time, It will be easier to review and test the changes, no guarantee you will get an answer from the maintainers  though.

GermanAizek:
@s1lentq I accidentally made a force push.
I hope you have pick up necessary fixes?
Added new commit with fix minor memleak https://github.com/dreamstalker/rehlds/pull/942/commits/1fa34c2645006f2023c5b215010f8a7f19153986

rehlds/HLTV/Core/src/BSPModel.h
```diff
@@ -45,7 +45,7 @@ class IBaseSystem;
 class BSPModel: public IBSPModel {
 public:
 	BSPModel() {}
-	virtual ~BSPModel() {}
+	virtual ~BSPModel() { delete m_wadpath; }
```
s1lentq, line 48: 'm_wadpath` should be released by `Free` function instead of `delete` operator, because the pointer was allocated using `_strdup`

PR #952, opened 2023-02-18
Add a new clip type that allows to trace when ShouldCollide is false

justgo97:
When we return 0 in pfnShouldCollide it can be useful to use this for RadiusFlash_TraceLine, FireBullets, IsVisible... etc


hajimura:
The feature should be implemented in regamedll too

justgo97:
> The feature should be implemented in regamedll too

Yes, If this gets to add to ReHLDS then it's easy to change the trace type in ReGameDLL for the relevant functions

justgo97:
closed in favor of https://github.com/s1lentq/ReGameDLL_CS/pull/813

PR #976, opened 2023-07-01
Implement -pingboost 4 for behavior similar to mm_insane 1 in Linux

cris840:
Implemented the Sleep_Pingboost4 function to achieve similar behavior to mm_insane 1 of the mmtimer metamod, since the function does not perform any further action, sleep is removed.

### DIFF
--- a/rehlds/dedicated/src/sys_linux.cpp
+++ b/rehlds/dedicated/src/sys_linux.cpp
@@ -146,34 +146,40 @@ void alarmFunc(int num)
 
 }
 
+void Sleep_Pingboost4(int msec);
+
 void Sys_InitPingboost()
 {
-	Sys_Sleep = Sleep_Old;
+    Sys_Sleep = Sleep_Old;
 
-	char *pPingType;
-	if (CommandLine()->CheckParm("-pingboost", &pPingType) && pPingType) {
-		int type = Q_atoi(pPingType);
-		switch (type) {
-		case 1:
-			signal(SIGALRM, alarmFunc);
-			Sys_Sleep = Sleep_Timer;
-			break;
-		case 2:
-			Sys_Sleep = Sleep_Select;
-			break;
-		case 3:
-			Sys_Sleep = Sleep_Net;
+    char* pPingType;
+    if (CommandLine()->CheckParm("-pingboost", &pPingType) && pPingType) {
+        int type = Q_atoi(pPingType);
+        switch (type) {
+        case 1:
+            signal(SIGALRM, alarmFunc);
+            Sys_Sleep = Sleep_Timer;
+            break;
+        case 2:
+            Sys_Sleep = Sleep_Select;
+            break;
+        case 3:
+            Sys_Sleep = Sleep_Net;
+            NET_Sleep_Timeout = (NET_Sleep_t)Sys_GetProcAddress(g_pEngineModule, "NET_Sleep_Timeout");
+            break;
+        case 4: // New case for -pingboost 4
+            Sys_Sleep = Sleep_Pingboost4;
+            break;
+        default:
+            Sys_Sleep = Sleep_Old;
+            break;
+        }
+    }
+}
 
-			// we Sys_GetProcAddress NET_Sleep() from
-			//engine_i486.so later in this function
-			NET_Sleep_Timeout = (NET_Sleep_t)Sys_GetProcAddress(g_pEngineModule, "NET_Sleep_Timeout");
-			break;
-		// just in case
-		default:
-			Sys_Sleep = Sleep_Old;
-			break;
-		}
-	}
+void Sleep_Pingboost4(int msec)
+{
+    // Do nothing, just return without sleep
 }
 
 void Sys_WriteProcessIdFile()


rehlds/dedicated/src/sys_linux.cpp
```diff
@@ -146,34 +146,40 @@ void alarmFunc(int num)
...
+        case 1:
+            signal(SIGALRM, alarmFunc);
+            Sys_Sleep = Sleep_Timer;
+            break;
+        case 2:
+            Sys_Sleep = Sleep_Select;
+            break;
+        case 3:
+            Sys_Sleep = Sleep_Net;
+            NET_Sleep_Timeout = (NET_Sleep_t)Sys_GetProcAddress(g_pEngineModule, "NET_Sleep_Timeout");
```
justgo97, line 168: Bring back the comments

rehlds/dedicated/src/sys_linux.cpp
```diff
@@ -146,34 +146,40 @@ void alarmFunc(int num)
...
+            Sys_Sleep = Sleep_Pingboost4;
+            break;
+        default:
+            Sys_Sleep = Sleep_Old;
+            break;
+        }
+    }
+}
+
+void Sleep_Pingboost4(int msec)
```
justgo97, line 180: Move this above so you can use it directly

cris840 (commented):
Alright

axl303:
Sorry for my stupid comment/question,
What would be the benefits/is there any from -pingboost4 and should we use it over -pingboost3?

cris840:
> Sorry for my stupid comment/question, What would be the benefits/is there any from -pingboost4 and should we use it over -pingboost3?

It's simple, in short:

-pingboost 3 is designed to control Sleep on network operations with precision while the network subsystem does its internal tasks.
 
-pingboost 4 does not perform any Sleep and simply returns the function without doing anything. This means that the CPU will be fully utilized without waiting or suspending execution. This can be used in order to increase FPS to "insane" levels ;)
Like mmtimer Metamod

SmilexGamer:
> Implemented the Sleep_Pingboost4 function to achieve similar behavior to mm_insane 1 of the mmtimer metamod, since the function does not perform any further action, sleep is removed.

Couldn't this be implemented also in Windows?

cris840:
> > Implemented the Sleep_Pingboost4 function to achieve similar behavior to mm_insane 1 of the mmtimer metamod, since the function does not perform any further action, sleep is removed.
> 
> Couldn't this be implemented also in Windows?

On Windows you can use the mmtimer metamod directly and use mm_insane 1 in server.cfg to have the same result.

SmilexGamer:
> > > Implemented the Sleep_Pingboost4 function to achieve similar behavior to mm_insane 1 of the mmtimer metamod, since the function does not perform any further action, sleep is removed.
> > 
> > 
> > Couldn't this be implemented also in Windows?
> 
> On Windows you can use the mmtimer metamod directly and use mm_insane 1 in server.cfg to have the same result.

Doesn't that apply also to Linux? You can use the mmtimer too. If we're implementing this in reHLDS, it should be in both platforms.

cris840:
> 



> > > > Implementé la función Sleep_Pingboost4 para lograr un comportamiento similar a mm_insane 1 del metamod mmtimer, ya que la función no realiza ninguna otra acción, se elimina el sueño.
> > > 
> > > 
> > > ¿No se podría implementar esto también en Windows?
> > 
> > 
> > En Windows, puede usar el metamod mmtimer directamente y usar mm_insane 1 en server.cfg para obtener el mismo resultado.
> 
> ¿Eso no se aplica también a Linux? También puede utilizar el mmtimer. Si estamos implementando esto en reHLDS, debería estar en ambas plataformas.



> > > > Implemented the Sleep_Pingboost4 function to achieve similar behavior to mm_insane 1 of the mmtimer metamod, since the function does not perform any further action, sleep is removed.
> > > 
> > > 
> > > Couldn't this be implemented also in Windows?
> > 
> > 
> > On Windows you can use the mmtimer metamod directly and use mm_insane 1 in server.cfg to have the same result.
> 
> Doesn't that apply also to Linux? You can use the mmtimer too. If we're implementing this in reHLDS, it should be in both platforms.

pingboost does not exist in the Windows version of ReHLDS
Look: https://github.com/dreamstalker/rehlds/blob/master/rehlds/dedicated/src/sys_window.cpp

cris840:
> > > > Implemented the Sleep_Pingboost4 function to achieve similar behavior to mm_insane 1 of the mmtimer metamod, since the function does not perform any further action, sleep is removed.
> > > 
> > > 
> > > Couldn't this be implemented also in Windows?
> > 
> > 
> > On Windows you can use the mmtimer metamod directly and use mm_insane 1 in server.cfg to have the same result.
> 
> Doesn't that apply also to Linux? You can use the mmtimer too. If we're implementing this in reHLDS, it should be in both platforms.

And no, unfortunately there is no working version of mmtimer for Linux. It would need to be updated, but there is no point in doing so when it can be implemented directly in ReHLDS.

stambeto2006:
@cris840 It would be a good idea and useful if it was built to work on Windows as well without adding unnecessary extras that are not guaranteed to work 100%.

hun1er:
@stambeto2006 , @SmilexGamer You can use [this](https://github.com/hun1er/rehlds-m/releases/tag/v1.3.0) which is compatible with HLDS and ReHLDS and works on both Windows and Linux. It has a similar feature (pingboost 5) and even more.

rtxa:
@hun1er Wouldn't be better to do a PR with your changes instead of using your fork? Otherwise it always get behinds with latest ReHLDS.

anzz1:
This doesn't really make any sense as the max packet rate is 100/s anyway and already with the original method the server's tickrate is way above that.
What this does is completely unnecessarily waste CPU cycles and energy and could cause server providers to not adopt ReHLDS or even outright ban it.

SmilexGamer:
> Yes of course it makes difference, to a point. This however, makes no sense. As long as the server can process the frame in time for the next packet, the rest is pointless. In fact the absolute minimum frametime is 1ms anyway so anything beyond that will do absolutely nothing but waste cycles.

That's odd, since I've seen a difference past 1ms frametime. Feel free to do the same test using ReHLDS-M.

PR #951, opened 2023-02-09
Add SV_AllowPhysent hook

justgo97:
Can be used to create a custom semiclip without needing a module

Ex:

`#include <amxmodx>
#include <cstrike>
#include <fakemeta>
#include <reapi>

public plugin_init() {
    register_plugin("Semiclip", "0.0.1", "JustGo")
    
    register_forward(FM_AddToFullPack,"fw_AddToFullPack_Post",1)

    RegisterHookChain(RH_SV_AllowPhysent, "SV_AllowPhysent")
}

public SV_AllowPhysent(ent, sv_player) {
    if(!is_user_alive(ent) || !is_user_alive(sv_player)) return HC_CONTINUE

    if(!isSameTeam(ent, sv_player)) return HC_CONTINUE

    SetHookChainReturn(ATYPE_BOOL, false)
    return HC_SUPERCEDE
}

public fw_AddToFullPack_Post(es_handle,e,ent,host,hostflags,player,pSet)
{
    if(!is_user_alive(host) || !is_user_alive(ent)) return FMRES_IGNORED

    if(isSameTeam(host, ent))
    {
        set_es(es_handle, ES_Solid, SOLID_NOT);
    }
    return FMRES_HANDLED;
}

stock isSameTeam(id1, id2) {
    return (cs_get_user_team(id1) == cs_get_user_team(id2))
}
`

PR #980, opened 2023-08-19
Add GetDispatch hook

tmp64:
This PR adds a hook for `GetDispatch` function. It is used by [WeaponMod](https://github.com/tmp64/weaponmod/).

It additionally bumps the API version to 3.14~~159265~~. #951 also bumped the version in `rehlds/version/version.h` but I didn't. Let me know and I'll fix the commit.

### DIFF
--- a/rehlds/engine/sys_dll.cpp
+++ b/rehlds/engine/sys_dll.cpp
@@ -678,6 +678,11 @@ qboolean EXT_FUNC Voice_SetClientListening(int iReceiver, int iSender, qboolean 
 
 DISPATCHFUNCTION GetDispatch(char *pname)
 {
+	return g_RehldsHookchains.m_GetDispatch.callChain(GetDispatch_Internal, pname);
+}
+
+DISPATCHFUNCTION GetDispatch_Internal(char* pname)
+{
 	int i;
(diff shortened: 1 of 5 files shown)


StevenKal:
Despite having "GetDispatch" hook could be better (since this is the "inside" function), was not the already existing hook "GetEntityInit" not enough for WeaponMod?

tmp64:
You're right. I could just hook `GetEntityInit` since it simply calls `GetDispatch`. I'll check and report back later

```cpp
ENTITYINIT GetEntityInit_internal(char *pClassName)
{
	return (ENTITYINIT)GetDispatch(pClassName);
}

ENTITYINIT EXT_FUNC GetEntityInit_api(char *pClassName)
{
	return g_RehldsHookchains.m_GetEntityInit.callChain(GetEntityInit_internal, pClassName);
}

ENTITYINIT GetEntityInit(char *pClassName)
{
	return GetEntityInit_api(pClassName);
}
```

tmp64:
Yep, `GetEntityInit` hook works just fine. `GetDispatch` seems to be used as a generic `dlsym` in the engine. I guess this PR could be left as is in case someone wants to hook GameDLL functions without a hooking library.

PR #987, opened 2023-10-02
Avoid redundant ping updates

justgo97:
Every client ping is recalculated with a delay of two seconds so it's redundant to update them every frame. I guess if too many players are pressing the IN_SCORE button on a server with sys_ticrate 1000+ It's gonna be a significant wasted bandwidth.


justgo97:
Closing this since I figured that there is more to it since the pings are calculated in an individualistic approach. though another kind of refactoring can be done.

justgo97:
To notice the bandwidth waste just put net_graph 3 and open the scoreboard in a full server, you will notice how the incoming rate spikes up, the higher the sys_ticrate of a server the higher the spike will be, that's solved with this PR. This will probably help players who experience an overflow/lag to not make things worse when they open their scoreboard.

PR #993, opened 2023-11-05
Add command line option "-map" to load entity file

golukon:
When you launch rehlds with "+map" command, entity file "*.ent" don't load. The problem occurs because the map is loaded first, then the cvars. New command line allows you to load cvars first


s1lentq:
@golukon Do you mean about CVar sv_use_entity_file?
Why not use `+sv_use_entity_file 1` before `+map` ?

golukon:
It really works, so apparently I can close PR

PR #994, opened 2023-11-11
Implemented reduction of impact caused by zip-bomb exploit

s1lentq:
This PR enhance security by controlling the decompression of incoming network data buffer to avoid server hangs and another potential risks associated with zip bomb attacks.

Added network security CVars:
- `sv_net_incoming_decompression` (0-1) Enables or disables incoming data decompression. Default: 1
- `sv_net_incoming_decompression_max_ratio` (0.0 - 100.0) Sets max allowed ratio between compressed and decompressed data. (A ratio close to 90 indicates large uncompressed data with low entropy, this means abnormal compressed data which can cause server to hang). Default: 80
- `sv_net_incoming_decompression_max_size` (16-65536) Adjusts max size in bytes of output data after decompression. Default: 65536
- `sv_net_incoming_decompression_punish` Time in minutes for which the player will be banned for malformed/abnormal bzip2 fragments (0 - Permanent, use a negative number for a kick). Default: -1

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,9 @@ This means that plugins that do binary code analysis (Orpheu for example) probab
 <li>sv_rehlds_local_gametime <1|0> // A feature of local gametime which decrease "lags" if you run same map for a long time. Default: 0
 <li>sv_use_entity_file // Use custom entity file for a map. Path to an entity file will be "maps/[map name].ent". 0 - use original entities. 1 - use .ent files from maps directory. 2 - use .ent files from maps directory and create new .ent file if not exist.
 <li>sv_usercmd_custom_random_seed // When enabled server will populate an additional random seed independent of the client. Default: 0
+<li>sv_net_incoming_decompression <1|0> // When enabled server will decompress of incoming compressed file transfer payloads. Default: 1
+<li>sv_net_incoming_decompression_max_ratio <0|100> // Sets the max allowed ratio between compressed and uncompressed data for file transfer. (A ratio close to 90 indicates large uncompressed data with low entropy) Default: 80.0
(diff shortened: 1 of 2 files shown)


fred0r:
`sv_net_incoming_decompression_punish` doesnt return any value in serverconsole?!

fred0r:
thx

fred0r:
so when `sv_allowupload 0`, `sv_net_incoming_decompression` doesnt matter?

drag1c:
Tested, both players were using GS Client, both got kicked with message Malformed/abnormal compressed data.
Used default values for cvars.

s1lentq:
> Tested, both players were using GS Client, both got kicked with message Malformed/abnormal compressed data. Used default values for cvars.

Try raise value of the cvar `sv_net_incoming_decompression_max_ratio` by step 0.1 until false positives is gone

s1lentq:
> so when `sv_allowupload 0`, `sv_net_incoming_*` doesnt matter?

no, sv_allowupload allows you to upload files fragment, but this PR fixes only normal fragment

drag1c:
> > Tested, both players were using GS Client, both got kicked with message Malformed/abnormal compressed data. Used default values for cvars.
> 
> Try raise value of the cvar `sv_net_incoming_decompression_max_ratio` by step +1.0 until false positives is gone
> 
> p.s can you capture incoming traffic into pcap format from these clients and share it to my mail?

a bit challenging but I will try, just to catch guys who had issue.

RauliTop:
@s1lentq 
Just write you to clarify about this.
The information here exposed is probably too much technical. We need more simple explanation.

When the decompression is being used?
I mean, what clients use it. For example, on GSClient, is working without decompression too?

How about if I disable the main cvar. Can players enter in the server?

PR #1000, opened 2023-11-20
engine: add sv_allow_autoaim cvar for HL25 DLL compatibility

a1batross:
Allows to work with HL25 server DLL, which acquires this new cvar and doesn't check against `nullptr`.

Was adding HL25 compatibility to Xash3D FWGS and decided to send this patch here too.


rndtrash:
damn, i was about to make a pr :(
lgtm!

PR #1002, opened 2023-11-21
Increase engine limits in accordance to the new update

fantoro:
Increase packet entity limit for compatibility with the 25th anniversary update's maps (Primarily rocket_frenzy, at least that's the only one that caught my attention so far) in accordance to the limit present in the 25th anniversary update's engine.

### DIFF
--- a/rehlds/common/qlimits.h
+++ b/rehlds/common/qlimits.h
@@ -39,6 +39,6 @@ constexpr auto MAX_LIGHTSTYLE_SIZE = size_t{64};
 #define MAX_BASE_DECALS			(1<<MAX_DECAL_INDEX_BITS)
 
 #define MAX_EVENTS			256
-#define MAX_PACKET_ENTITIES		256	// 256 visible entities per frame
+#define MAX_PACKET_ENTITIES		1024	// 1024 visible entities per frame
 
 #endif // QLIMITS_H
--- a/rehlds/engine/server.h
+++ b/rehlds/engine/server.h
@@ -31,7 +31,7 @@
 #include "maintypes.h"
 
 // TODO: I think this defines must be in /common/
-const int NUM_EDICTS = 900;
+const int NUM_EDICTS = 1200;
 const int MAX_NAME   = 32;
 
 #include "custom_int.h"


wopox1337:
Each mod overrides the num_edicts value in the `<gamedir>/liblist.gam` file

https://github.com/dreamstalker/rehlds/blob/ecfe8bb5fb7a8f82c13113ac10ef88461878d8fc/rehlds/engine/sys_dll.cpp#L837-L842

Perhaps this is what the developers of the update changed? (I haven't looked in the IDA).

rehlds/common/qlimits.h
```diff
@@ -39,6 +39,6 @@ constexpr auto MAX_LIGHTSTYLE_SIZE = size_t{64};
 #define MAX_BASE_DECALS			(1<<MAX_DECAL_INDEX_BITS)
 
 #define MAX_EVENTS			256
-#define MAX_PACKET_ENTITIES		256	// 256 visible entities per frame
```
wopox1337, line 42: At the moment, when game clients have not updated and have an old version, this change can be critical and older versions of game clients will get critical errors.

wopox1337:
Archiving until better times... (when Valve has fixed their updates)
image

StevenKal:
Need the engine be able to detect the client version (HL25...) & apply changes & limits in consequence.

dystopm:
> Need the engine be able to detect the client version (HL25...) & apply changes & limits in consequence.

In this case everybody should upgrade their clients (Steam, maybe GSClient dev team also) and ReHLDS too, we cannot stick with the older versions if we want better features and tools for modding

StevenKal:
When some recent clients are causing troubles because VALVe did not tested all well, or need deep changes in our client & server configs, etc., not everyone will like to update, neither have time or motivation.
For example on server configs, when offsets, signatures, functions format & cie are fucked up on new updates (also if this is frequently updated & refucked up later), a lot of people will just prefer to stick to their old servers binaries & configs rather than waste their time updating, for benefits they might not judge necessary. Maybe on HL updating could worth, but most HL1 users & players are on CS, & from a moment we have the chance to have Re* projects as example, VALVe is probably not gonna release an update 2x better than those, they have the possibility to upgrade "legally" both servers & clients binaries to make them tied automatically via updates, but not all worth.

I always thinked that tried to provide support for old versions, by detecting client's build is the best to avoid issues.
I did not tried neither checked their HL25 update, but did they added new CVar "game_version" or something like? I recall suggested such [here](https://github.com/ValveSoftware/halflife/issues/2913).

dystopm:
> When some recent clients are causing troubles because VALVe did not tested all well, or need deep changes in our client & server configs, etc., not everyone will like to update, neither have time or motivation for such, also knowing nowadays people on HL1 might no longer have same motivation than before. For example on server configs, when offsets, signatures, functions format & cie are fucked up on new updates (also if this is frequently updated & refucked up later), a lot of people will just prefer to stick to their old servers binaries & configs rather than waste their time updating, for benefits they might not judge necessary. Maybe on HL updating could worth, but most HL1 users & players are on CS, & from a moment we have the chance to have Re* projects as example, VALVe is probably not gonna release an update 2x better than those, they have the possibility to upgrade "legally" both servers & clients binaries to make them tied automatically via updates, but not all worth.
> 
> I always thinked that tried to provide support for old versions, by detecting client's build is the best to avoid issues. I did not tried neither checked their HL25 update, but did they added new CVar "game_version" or something like? I recall suggested such [here](https://github.com/ValveSoftware/halflife/issues/2913).

I agree with you. Thing is, if we want a way to make client show his version to the server engine, an update is requiered, so everybody should be forced to update their binaries. It's cyclical

StevenKal:
> I agree with you. Thing is, if we want a way to make client show his version to the server engine, an update is required, so everybody should be forced to update their binaries. It's cyclical

Current CVar "sv_version" on the client can be queried by the engine (& parsed...), or the presence of new CVar(s) added after each version could be used as second tricky workaround, but unfortunately a minor "ping delay" until he replies (so not accessible at first "server contact"), also the fact the client can alter it (but usually the client will not think to trick it), while devs should make it unmodifiable/locked.
Setinfo can worth too (like for "cl_lw"...), but currently its length of 256 is limited...

StevenKal:
The engine can obviously does it, but the value must be stored, so either add a new member on an existing or a new structure, either a new global variable.
Could be implement on "client_t" but this will break it (at least its size if put in the end), or easier, on "CGameClient".

dystopm:
> > Type "cvarlist log" on old build (on your client), then again on new build, use a text comparator like PSPad or else to check if there are new or removed CVars, etc., or query "sv_version" (& parse it), must be higher than 8684 for sure!
> 
> Is there a way for the server to query CVars on the client though? I figured there wasn't since it seems like the CVars that the server needs are put in `userinfo`.
> 
> EDIT: sorry for the stupid question, just found it lol

pfnCvarValue, but

> but unfortunately a minor "ping delay" until he replies (so not accessible at first "server contact")

dystopm:
> > pfnCvarValue, but
> > > but unfortunately a minor "ping delay" until he replies (so not accessible at first "server contact")
> 
> I think a slight delay is better than nothing atp tbh.
> 
> Also thank you!

We might wait to see an update from Valve first, this is breaking everything

StevenKal:
> I think a slight delay is better than nothing atp tbh.
> 
> Also thank you!

Yeah if we query it from client connection, we will have the value before it enters the game (pfnPutInServer), so this should work fine for our need & cover most of the areas where only in-game codes are used.
But yeah, better wait for a more stable/definitive update by VALVe, but we do not know what are their plans, if this is just a minor one for 25th, or if they magically plan to update all HL1 games in a mass with a bunch of changes & features.

mlibre2:
apparently this time they are doing "[private](https://steamdb.info/app/10/history/)" tests before releasing future updates

FEDERICOMB96:
CS and CS:CZ updated 2h ago to support the HL25th engine updates. Finally.
https://store.steampowered.com/news/app/10/view/3889483908318472024?l=latam

I understand that implementing these changes will make clients with older versions incompatible. However, I also understand that it is necessary if we want better features and tools for modding. I would have liked them to expand the limits of the sounds/models/sprites/etc precache as well. Perhaps they forgot (?

PR #1014, opened 2024-01-19
SZ_GetSpace: Fix reversing error & buffer overflow

anzz1:
**edit:** tl;dr; as it becomes clear in the posts further below, this PR doesn't actually fix the problem. 

---

SZ_GetSpace: Fix reversing error & buffer overflow

This hopefully fixes the crash following soon after `SZ_GetSpace: overflow on netchan->message` and other SZ_GetSpace related errors.

Also I have a question about previous implementation,
where did these come from?

https://github.com/dreamstalker/rehlds/blob/90cb97ddae6cd7c66b08987d4d8049cb9fbcc76d/rehlds/engine/common.cpp#L1189-L1192

and

https://github.com/dreamstalker/rehlds/blob/90cb97ddae6cd7c66b08987d4d8049cb9fbcc76d/rehlds/engine/common.cpp#L1243

These are not present in the original engine, nor are they wrapped in `#ifdef REHLDS_FIXES`. It seems they appear out of thin air.

### DIFF
--- a/rehlds/engine/common.cpp
+++ b/rehlds/engine/common.cpp
@@ -1182,72 +1182,45 @@ qboolean SZ_HasSomethingToRead(const sizebuf_t *buf, int length)
 
 void *EXT_FUNC SZ_GetSpace(sizebuf_t *buf, int length)
 {
-	void *data;
+	unsigned char *data;
 	const char *buffername = buf->buffername ? buf->buffername : "???";
+	int cursize = buf->cursize;
 
-
-	if (length < 0)
+	if (cursize + length <= buf->maxsize)
 	{
-		Sys_Error("%s: %i negative length on %s", __func__, length, buffername);
+		data = buf->data;
+		buf->cursize += length;
+		return &data[cursize];
 	}
 
-	if (buf->cursize + length > buf->maxsize)
+	if (!(buf->flags & SIZEBUF_ALLOW_OVERFLOW))
 	{
-#ifdef REHLDS_FIXES
+		if (!buf->maxsize)
+		{
+			Sys_Error("%s: tried to write to an uninitialized sizebuf_t: %s", __func__, buffername);
+		}
+		else
+		{
+			Sys_Error("%s: overflow without FSB_ALLOWOVERFLOW set on %s", __func__, buffername);
+		}
+	}
+
+	if (length > buf->maxsize)
+	{
 		if (!(buf->flags & SIZEBUF_ALLOW_OVERFLOW))
 		{
-			if (!buf->maxsize)
-			{
-				Sys_Error("%s: tried to write to an uninitialized sizebuf_t: %s", __func__, buffername);
-			}
-			else if (length > buf->maxsize)
-			{
-				Sys_Error("%s: %i is > full buffer size on %s", __func__, length, buffername);
-			}
-			else
-			{
-				Sys_Error("%s: overflow without FSB_ALLOWOVERFLOW set on %s", __func__, buffername);
-			}
+			Sys_Error("%s: %i is > full buffer size on %s", __func__, length, buffername);
 		}
-
-		if (length > buf->maxsize)
+		else
 		{
 			Con_DPrintf("%s: %i is > full buffer size on %s, ignoring", __func__, length, buffername);
 		}
-#else // REHLDS_FIXES
-		if (!(buf->flags & SIZEBUF_ALLOW_OVERFLOW))
-		{
-			if (!buf->maxsize)
-			{
-				Sys_Error("%s: Tried to write to an uninitialized sizebuf_t: %s", __func__, buffername);
-			}
-			else
-			{
-				Sys_Error("%s: overflow without FSB_ALLOWOVERFLOW set on %s", __func__, buffername);
-			}
-		}
-
-		if (length > buf->maxsize)
-		{
-			if (!(buf->flags & SIZEBUF_ALLOW_OVERFLOW))
-			{
-				Sys_Error("%s: %i is > full buffer size on %s", __func__, length, buffername);
-			}
-
-			Con_DPrintf("%s: %i is > full buffer size on %s, ignoring", __func__, length, buffername);
-		}
-#endif // REHLDS_FIXES
-
-		Con_Printf("%s: overflow on %s\n", __func__, buffername);
-
-		SZ_Clear(buf);
-		buf->flags |= SIZEBUF_OVERFLOWED;
 	}
 
-	data = &buf->data[buf->cursize];
-	buf->cursize = length + buf->cursize;
-
-	return data;
+	Con_Printf("%s: overflow on %s\n", __func__, buffername);
+	buf->cursize = length;
+	buf->flags |= SIZEBUF_OVERFLOWED;
+	return buf->data;
 }
 
 void SZ_Write(sizebuf_t *buf, const void *data, int length)


anzz1:
It's pretty clear when you look at the original func that it's wrong.

Not only it does not match the original `SZ_GetSpace` function, but it doesn't function correctly.

The return value is same whether the sizebuf overflowed or not. Only thing that changes between the normal and overflowed codepaths is the added `SZ_Clear` clearing all `buf->flags` except `SIZEBUF_OVERFLOWED`

SZ_Clear also sets `buf->cursize` to 0, but it doesn't matter as its overwritten on the next line.
https://github.com/dreamstalker/rehlds/blob/90cb97ddae6cd7c66b08987d4d8049cb9fbcc76d/rehlds/engine/common.cpp#L1248

So these lines
https://github.com/dreamstalker/rehlds/blob/90cb97ddae6cd7c66b08987d4d8049cb9fbcc76d/rehlds/engine/common.cpp#L1243-L1244
are exactly the same as just doing:
```c
buf->flags = SIZEBUF_OVERFLOWED
```

anzz1:
It's pretty clear when you look at the original func that it's wrong.

Not only it does not match the original `SZ_GetSpace` function, but it doesn't function correctly.

The return value is same whether the sizebuf overflowed or not.
Ultimately the only thing that changes between the normal and overflowed codepaths are the prints.

The changes that are different end up doing nothing:

SZ_Clear unsets the `SIZEBUF_OVERFLOWED` flag, which is directly set again afterwads. Nothing is done here.
https://github.com/dreamstalker/rehlds/blob/90cb97ddae6cd7c66b08987d4d8049cb9fbcc76d/rehlds/engine/common.cpp#L1243-L1244

SZ_Clear also sets `buf->cursize` to 0, but it doesn't matter as its overwritten on the next line.
https://github.com/dreamstalker/rehlds/blob/90cb97ddae6cd7c66b08987d4d8049cb9fbcc76d/rehlds/engine/common.cpp#L1248

s1lentq:
There is no reverse mistake here, and the code behavior is exactly the same, except for the lack of `REHLDS_FIXES` to check negative length

@anzz1 
You may be misled by looking at disassembly code as it, but when combined with DWARF and take into account the omitted code by optimizing the compiler, the code looks exactly as it should be

s1lentq:
> Secondly, the `SZ_Clear` function is never called anywhere in the `SZ_GetSpace` function, in any case.

You can check with dwarf again and use the right tools to fully analyze them, and you will notice that `SZ_Clear` was used in the vanilla code, and the fact that you don't see this function in the disassembly code means about compiler optimization, because GCC using more optimizations of compiler (like -O3, -ipo etc) by compared windows compiler VC++ (/O2 /Os /GL etc)

```c
/* <11cdc> ../engine/common.cpp:1777 */
void *SZ_GetSpace(sizebuf_t *buf, int length)
{
//	void *data;
//	SZ_Clear(sizebuf_t *buf);
}
```

If you still have doubts, u can take a look at windows binaries any build, which uses less optimization and cleaner disassembler (pseudocode) code

<details>
 <summary>SZ_GetSpace</summary>

```c
void *__cdecl SZ_GetSpace(sizebuf_t *buf, int length)
{
	bool v3; // zf
	const char *v4; // eax
	const char *v5; // eax
	const char *v6; // eax
	int v7; // ecx
	void *result; // eax

	if ( length + buf->cursize > buf->maxsize )
	{
		if ( !(buf->flags & 1) )
		{
			v4 = buf->buffername;
			if ( buf->maxsize == 0 )
			{
				if ( !v4 )
					v4 = "???";

				Sys_Error("SZ_GetSpace:	Tried to write to an uninitialized sizebuf_t: %s", v4);
			}

			if ( !v4 )
				v4 = "???";

			Sys_Error("SZ_GetSpace: overflow without FSB_ALLOWOVERFLOW set on %s", v4);
		}

		if ( length > buf->maxsize )
		{
			v5 = buf->buffername;
			if ( !(buf->flags & 1) )
			{
				if ( !v5 )
					v5 = "???";

				Sys_Error("SZ_GetSpace: %i is > full buffer size on %s", length, v5);
			}

			if ( !v5 )
				v5 = "???";

			Con_DPrintf("SZ_GetSpace: %i is > full buffer size on %s, ignoring", length, v5);
		}

		v6 = buf->buffername;
		if ( !buf->buffername )
			v6 = "???";

		Con_Printf("SZ_GetSpace: overflow on %s\n", v6);
		SZ_Clear(buf);
		buf->flags |= 2u;
	}
	v7 = buf->cursize;
	result = &buf->data[v7];
	buf->cursize = length + v7;
	return result;
}
```
</details>


Anyway, let's leave the point about what a function looks like in vanilla code and turn to the main question, the logic of the code.

> But currently it does not have the "If overflowed" part.
It always return the buf->data + cursize, even if overflowed.

No, is not

> If overflowed:
Set buf->cursize to length.
Add flag SIZEBUF_OVERFLOWED to buf->flags
Return pointer to buf->data

> If not overflowed:
Set buf->cursize to buf->cursize + length
Return pointer to buf->data + cursize (note: the previous cursize, not cursize+length !)

Following your point of view about the logic of the code, there are no contradictions in the current code, it still works as you expected, and for some reason, you ignore the fact that `cursize` is always 0 in overflow case after call SZ_Clear, therefore it is equivalent to returning `data = &buf->data[0];`

anzz1:
Oh right, I don't have any DWARF comments only the function names, types and so.

Yes, I only checked the Linux version, indeed you are correct that SZ_Clear is called on the Windows swds.dll version.

So the only part that is wrong is the "%s: %i is > full buffer size on %s" / "%s: %i is > full buffer size on %s, ignoring" part as only one of them must be called, but this shouldn't cause any problems. I don't know which decompiler you are using, but at least IDA and Hex-Rays C decompiler does produce incorrect result here when looking at the Windows binary, while the Linux one produces correct result. The produced output from swds.dll looks what you posted above, but is incorrect. If you look at the graph view and then text view, you see it lose the jmp instruction after Sys_Error call since it thinks its a noreturn function. This produces wrong graph and wrong pseudocode.

Meanwhile the graph & pseudocode on the linux one is correct, and it's either/neither, but not both.
image



So this bug unfortunately still eludes us. What is clear though is that after a SZ_GetSpace overflow happens, crash or other undefined behaviour will happen later. The cause of it is just somewhere else, not here.

anzz1:
I guess I could still reopen this since this way its more correct, with the Printf paths bifurcated.

s1lentq:
> If you look at the graph view and then text view, you see it lose the jmp instruction after Sys_Error call since it thinks its a noreturn function

Because, if the function `Sys_Error` with this attribute `noreturn` actually returns then UB (behavior is undefined)
Therefore the compilers, such as VC++ and GCC, may handle such scenarios differently and own discretion, and that's why leading to variations in code instructions.

I'm still convinced that it doesn't matter which instructions follow the `Sys_Error` call if the function, marked as noreturn, guarantees the exit point of the app.

anzz1:
Oh yeah, if I followed further, I would've seen that `Sys_Error` indeed is a noreturn function, instead of being errorneously marked as such.

God damn it, I thought that this issue would be finally solved.
I should've known that the fix couldn't have been this easy...

Exactly as you said, I was confused after all. :sweat_smile: 

Well, you live to learn, I guess.

lspublic:
Anzz1, this problem what you fixed for sz getspace where we can download beacuse i have problem in my server

PR #1031, opened 2024-05-13
Make the server sets Steamids to clients after the validation by Steam servers

Splatt581:
This pr fixes the Steam App Ownership Ticket hijacking/spoofing vulnerability and related exploits. Now the server will sets the steamid from the ticket only after the client connection has been validated by the Steam servers. Before this, the client will be assigned `STEAM_ID_PENDING` (sid 0).

In fact, the GoldSrc server followed similar logic before the Steamworks updates.

### DIFF
--- a/rehlds/engine/host.cpp
+++ b/rehlds/engine/host.cpp
@@ -475,6 +475,9 @@ void SV_DropClient_internal(client_t *cl, qboolean crash, const char *string)
 
 		if (!cl->fakeclient)
 			Netchan_Transmit(&cl->netchan, i, final);
+
+		cl->network_userid.m_SteamIDValidated = 0;
+		host_client->network_userid.m_SteamIDValidated = 0;
 	}
 
 	connection_time = realtime - cl->netchan.connect_time;
--- a/rehlds/engine/sv_main.cpp
+++ b/rehlds/engine/sv_main.cpp
@@ -8411,13 +8411,13 @@ char* EXT_FUNC SV_GetIDString_internal(USERID_t *id)
 		{
 			Q_strncpy(idstr, "STEAM_ID_LAN", ARRAYSIZE(idstr) - 1);
 		}
-		else if (!id->m_SteamID)
+		else if (!id->m_SteamIDValidated)
 		{
 			Q_strncpy(idstr, "STEAM_ID_PENDING", ARRAYSIZE(idstr) - 1);
 		}
 		else
 		{
-			TSteamGlobalUserID steam2ID = Steam_Steam3IDtoSteam2(id->m_SteamID);
+			TSteamGlobalUserID steam2ID = Steam_Steam3IDtoSteam2(id->m_SteamIDValidated);
 			Q_snprintf(idstr, ARRAYSIZE(idstr) - 1, "STEAM_%u:%u:%u", steam2ID.m_SteamInstanceID, steam2ID.m_SteamLocalUserID.Split.High32bits, steam2ID.m_SteamLocalUserID.Split.Low32bits);
 		}
 		break;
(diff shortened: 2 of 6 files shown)


Splatt581:
In a good way, we also needs to replace the `SendUserConnectAndAuthenticate` function with `BeginAuthSession`, which using in Source1/S2 engines. The `SendUserConnectAndAuthenticate` function is deprecated in the Steamworks SDK and has poor client auth ticket fields validation.

But for now in the local rebuild of rehlds the `BeginAuthSession` function returns `k_EBeginAuthSessionResultInvalidTicket` even with a valid auth ticket for appid 10.

tupec:
You can't spoof someone's ticket, there's a reason why - it's RSA signed with a private key. You can only reuse them (until the expiration, which is 21 days).

Splatt581:
> You can't spoof someone's ticket, there's a reason why - it's RSA signed with a private key. You can only reuse them (until the expiration, which is 21 days).

That's right.

But in the future, attackers may obtain the private key or otherwise start signing tickets. GoldSrc game servers must be ready for this.

“Reusing” of a ticket not with the owner’s account has been the main exploit used by attackers for 10 years; by stealing a ticket from server admins (mostly using social engineering), they gain admin rights on the server.

There are also other exploits that allow you to hide your steam avatar in the scoreboard, as well as a complete bypass of the steamid ban via the `banid` engine command.

This pr fixes all currently known vulnerabilities with steam tickets on the server.

anzz1:
Would this be better as a rehlds plugin?

Hooking the functions with the hookchain API and changing the result?
If not, at very least all this should be wrapped in `#ifdef REHLDS_FIXES` like all functionality that doesn't match original HLDS.

I could see potential compatibility issue with plugins, let's say you're running a zombie mod which does stuff on client connect and expects the user to have a id instead of STEAM_ID_PENDING.

It seems that `CSteam3Server::OnGSClientApprove()` which validates the connection doesn't call any rehlds hookchain or isn't connected to Metamod either, but it does seem that AMX Mod X still has support for this from pre-steamworks era and should only call [client_authorized()](https://www.amxmodx.org/api/amxmodx/client_authorized) after the id is actually validated.

On [ClientConnect](https://github.com/alliedmodders/amxmodx/blob/27f451a868c3154a0ab0d4d70e0a650074fd485d/amxmodx/meta_api.cpp#L877)(post) it does only call amxx's client_authorized() if the steam id is not STEAM_ID_PENDING, and if it is adds the player to a list which is then iterated in [StartFrame](https://github.com/alliedmodders/amxmodx/blob/master/amxmodx/meta_api.cpp#1219)(post) and client_authorized() is called and the player removed from list if he has got the validation since.
That all is from a quick glance though, no idea if it actually still works or could have been broken since considering we've already had Steamworks for a small eternity.

This change would no longer return valid id in [client_connect()](https://www.amxmodx.org/api/amxmodx/client_connect) and the order of [client_putinserver()](https://www.amxmodx.org/api/amxmodx/client_putinserver) and [client_authorized()](https://www.amxmodx.org/api/amxmodx/client_authorized) could now happen in any order instead of the current client_connect=client_authorized, and client_putinserver afterwards.
Making a simple code change to affected plugins so that whatever player id related stuffs like loading loadouts or whatever would happen in client_putinserver() or client_authorized() whichever gets called later, would then "guarantee" (unless theres a hickup in the validation server) a valid steamid.

In any case changes like this can't be just pushed through like this, they should be definitely at least in `REHLDS_FIXES` considering this changes original functionality, or as it breaks plugin compatibility maybe introduced as a plugin or wrapped in a cvar like `sv_rehlds_id_extra_validation`.

Splatt581:
@anzz1 

Thanks for the detailed feedback!

I looked in the `client_authorized` forward - unfortunately, at the moment it does not depend on the `CSteam3Server::OnGSClientApprove` call, so with the current functionality, we can't do without additional hooks in the plugin.

I agree, the current implementation of the fix is ​​pretty bad and done in a hurry. I like the idea of ​​an additional cvar and `REHLDS_FIXES` macroses, so I will do code reconstruction when I have time for it.

Issue #862, opened 2021-10-03
Server doesn't show up in server list when is paused


rtxa:
# Description

Basically when the server has been paused (by issuing the command `pause` and `pausable` has been enabled), it dissapears from **server list**, in other words it shows `< not responding >`.

Apparently, this is not true as you can still enter the server manually by using the command `connect` and the server IP.

# Expected behaviour

The expected behaviour would be that the server should appear in the server list even when is paused, **just like normal HLDS does.**

# Why is important to fix it

The `pause` command is a key feature for match competition in Half-Life and CS 1.6, and mods like Adrenaline Gamer. When a match has started and a player has been disconnect from the server by some reason like game crash or lost his internet connection, the remaining players will pause the server and wait for him. When the player has returned, they will unpause the server and continue with the match. But if the server **doesn't show up in the server list**, it can be misleading for the player as he can think that the server doesn't work and probably it will end up with aborting the match.

# What could be the reason

I'm not familiarized with ReHLDS code but perhaps the issue has something to do with the server not sending more responses to the client when it request server info on server pause. I'll appreciate if someone can fix it or help to lead to the fix.

culcj95:
> Basically when the server is paused (by using the command `pause` and `pausable` is enabled) it dissapears from **server list**, in other words it shows `< not responding >`.
> 
> This can be workaround by entering the server with the console by using the command `connect`. Take into account that this issue **doesn't happen in normal HLDS.**
> 
> The `pause` command is very used in matchs (1v1, 2v2 or more) when one of the players gets disconnect by reasons like game crash or lost of internet connection, requiring to pause the match until player comesback. But if the server doesn't show up in **server list**, it can be misleading for the player and he will think that the server doesn't work when it actually does.
> 
> What is expected is that server should still be displayed even when is paused, just like normal HLDS does.
> 
> I'm not familiarezed with HLDS code but perhaps the issue is with the server not sending more responses to the client when it request server info or something like that on pause. I'll appreciate if someone can fix this or help to guide to the fix.



DID YOU FIND ANY FIX FOR THIS???? @rtxa

culcj95:
@dreamstalker Any fix for this??

rtxa:
@culcj95 By the moment, I haven't found the solution. But I have updated the issue to be more descriptive and easy to understand.

abecee:
> @culcj95 By the moment, I haven't found the solution. But I have updated the issue to be more descriptive and easy to understand.

[...]

PR #1039, opened 2024-07-18
Implement API function "Host_Pause"

jonathan-up:
If need to pause, do not need to set `pausable` to `1` and then make client execute `pause` command anymore.
This stuff make pause easier and fix some clients may block `pause` commands sent by the server.

### DIFF
--- a/rehlds/engine/host_cmd.cpp
+++ b/rehlds/engine/host_cmd.cpp
@@ -2808,6 +2808,26 @@ void Host_Kill_f(void)
 	gEntityInterface.pfnClientKill(sv_player);
 }
 
+void EXT_FUNC Host_Pause_api(bool setPause)
+{
+	g_psv.paused = setPause;
+#ifdef REHLDS_FIXES
+	for (int i = 0; i < g_psvs.maxclients; i++)
+	{
+		if (g_psvs.clients[i].fakeclient)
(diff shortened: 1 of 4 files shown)


rtxa:
Thanks for implementing this, it would benefit AG and AG Mod X because (in my case) I rely on a random user to pause/unpause the server, which is odd and unrealiable. Maybe a bit offtopic, but did you take a look to this issue? https://github.com/dreamstalker/rehlds/issues/862 Unfortunately ReHLDS stops sending server info when on pause and makes players think the server is down, which is not the case in HLDS.

jonathan-up:
> but did you take a look to this issue? #862 
@rtxa I reply you in detail in that issues

anzz1:
Two things are missing here:

1) When you modify the API, you need to increase the minor API version in [rehlds_api.h](https://github.com/rehlds/ReHLDS/blob/master/rehlds/public/rehlds/rehlds_api.h)
2) When you add a new function to the `struct RehldsFuncs_t`, you need to add it to the **bottom of the struct** otherwise it breaks every plugin compiled with earlier minor API version.

Check https://github.com/rehlds/ReHLDS/commit/084bd023d287909c634774d8e30a5f5d4f1b4e0c for reference

PR #1063, opened 2024-11-10
Fix zip bomb exploit for decompressing file fragments

Splatt581:
At this point, an attacker can still attack the server with a zip bomb using file fragments. Video proof:
[https://www.youtube.com/watch?v=QWHxdTwzZVM](https://www.youtube.com/watch?v=QWHxdTwzZVM)

This pr fixes this vulnerability.

### DIFF
--- a/rehlds/engine/net_chan.cpp
+++ b/rehlds/engine/net_chan.cpp
@@ -1596,8 +1596,31 @@ qboolean Netchan_CopyFileFragments(netchan_t *chan)
 
 	Q_strncpy(compressor, MSG_ReadString(), sizeof(compressor) - 1);
 	compressor[sizeof(compressor) - 1] = 0;
+
+	qboolean success = TRUE;
+
 	if (!Q_stricmp(compressor, "bz2"))
+	{
+
+#ifdef REHLDS_FIXES
+		if (!sv_net_incoming_decompression.value)
+		{
+			if (chan->player_slot == 0)
+			{
+				Con_DPrintf("Incoming compressed data disallowed from\n");
+				return FALSE;
+			}
+			// compressed data is expected only after requesting resource list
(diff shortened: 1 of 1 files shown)


wopox1337:
You did something wrong with the commits, the changes are not displaying correctly, could you please fix it?

Splatt581:
Seems its done.

Garey27:
Can you also add check for file size that transmited? For example client is dropped with customization i attached.

[tempdecal.zip](https://github.com/user-attachments/files/17787430/tempdecal.zip)

Splatt581:
> Can you also add check for uncompressed file size that transmitted? For example client is dropped with customization i attached.
> 
> [tempdecal.zip](https://github.com/user-attachments/files/17787430/tempdecal.zip)

Yeah, I also recently encountered custom logos with max decompression ratios that exceed the allowed values. For example, your logo has max ratio of `85.91`. So far I see the following solutions:

1. Increasing the `sv_net_incoming_decompression_max_ratio` cvar to `90`. Of course, this may make the server more vulnerable to a zip bomb attack, I haven't tested it yet.

2. Maybe make max ratio for file fragments a separate cvar, something like `sv_file_incoming_decompression_max_ratio`? Because there is much more scope for manipulating file contents than with regular network messages.

3. Maybe we should introduce a warning system, something like `sv_net_incoming_decompression_max_warns`? To skip the firsts false-positive compressed data that the client sends (basically, as far as I remember, the client compresses only the logo decal of the file before sending, it doesn't even compress `clc_fileconsistency`).

What do you think about this?

Splatt581:
Fixed in this commit https://github.com/rehlds/ReHLDS/commit/64c684af4a8e6030b9e06a59a774d6efd92db352

PR #1074, opened 2025-01-05
Fix arbitrary network data sending exploit via malformed userinfo

Splatt581:
This is a new exploit that was first mentioned in this issue - https://github.com/rehlds/ReHLDS/issues/1073

By using invalid utf8 chars in userinfo, an attacker can interrupt the reading of the string in `svc_updateuserinfo` for all connected clients, which will likely cause them to disconnect from the server due to incorrect reading of the rest of the packet. However, in addition to simply disconnecting clients, an attacker can append arbitrary network traffic to his userinfo, which connected clients can process. For example, in the PoC video below, an attacker compromises the `svc_stufftext` message, which executes some commands/cvars in the console of other clients:

https://www.youtube.com/watch?v=BIG7I859_aI

The thing is that when processing userinfo via the `setinfo` command, the server already has a check in the code for invalid utf8 chars, it is located in the `Info_SetValueForStarKey` func:

```c
if (!Q_UnicodeValidate(key) || !Q_UnicodeValidate(value))
{
	Con_Printf("Keys and values must be valid utf8 text\n");
	return FALSE;
}
```

But there is no such check for userinfo sent via `C2S_CONNECTION` message, so I suggest adding it to `Info_IsValid` func.

So this pr fixes this vulnerability.

### DIFF
--- a/rehlds/engine/info.cpp
+++ b/rehlds/engine/info.cpp
@@ -856,6 +856,12 @@ qboolean Info_IsValid(const char *s)
 		return false;
 	};
 
+	// invalid utf8 chars are deprecated
+	if (!Q_UnicodeValidate(s))
+	{
+		return FALSE;
+	}
+
 	while (*s == '\\')
 	{
 		const char* key = ++s;


dystopm:
gj sir

UnrealKaraulov:
Yes 
image
image

```
topcolor\0\bottomcolor\0\cl_dlmax\1024\cl_lc\0\cl_lw\1\cl_updaterate\102\rate\100000\1\?cl_updaterateep\        name ~0WN3D~BY~X64CR3W~;spk hgrunt/c2a3_hg_laugh?\1\name\Player\model\urban\*sid\76561199447977504
topcolor\0\bottomcolor\0\cl_dlmax\1024\cl_lc\0\cl_lw\1\cl_updaterate\102\rate\100000\1\?cl_updaterateep\        name ~0WN3D~BY~X64CR3W~;spk hgrunt/c2a3_hg_laugh?\1\name\Player\model\urban\*sid\76561199447977504
```

UnrealKaraulov:
Now this exploit detected by https://github.com/UnrealKaraulov/UnrealDemoScanner
image

Xelson:
here is the attack traffic recorded on the demo:
image
There are two notable points here:
1) instead of the next key value pair, `0xFF` bytes are written, which stops parsing in `MSG_ReadStringLine`. A truncated userinfo string literally comes to the validation check.
https://github.com/rehlds/ReHLDS/blob/e9991716fc908e8b2ebf8ef5004c2a6f098f6eb5/rehlds/engine/common.cpp#L1092-L1097
called from here
https://github.com/rehlds/ReHLDS/blob/e9991716fc908e8b2ebf8ef5004c2a6f098f6eb5/rehlds/engine/sv_main.cpp#L3693-L3700
Technically, the string is valid, but it ends unexpectedly for the validator. At the same time, malicious traffic is recorded in subsequent bytes, which is decoded as `SVC_DISCONNECT` (`0x02`) with the corresponding message about the vacban
image
If such traffic had reached the player's netchan, it would have been perceived by the game client as a message about the player's kick for a vac ban, which actually did not happen.
2) despite the fact that the string itself is validated and truncated from the server's point of view, the players are not sent the string, but directly traffic from attacker's netchan
https://github.com/rehlds/ReHLDS/blob/e9991716fc908e8b2ebf8ef5004c2a6f098f6eb5/rehlds/engine/sv_main.cpp#L4088
and the problem here is that the string is read and sent exactly until the null terminator is detected, although `MSG_ReadStringLine` also reacts to `0xFF`
https://github.com/rehlds/ReHLDS/blob/e9991716fc908e8b2ebf8ef5004c2a6f098f6eb5/rehlds/engine/sv_main.cpp#L4064-L4067
https://github.com/rehlds/ReHLDS/blob/e9991716fc908e8b2ebf8ef5004c2a6f098f6eb5/rehlds/engine/common.cpp#L293-L297
as a result, it turns out that all players will receive the player's userinfo much longer than the server parsed. The remaining part of the string will be perceived by the client as part of the traffic from the server and will be executed as engine/user messages.

ghouls96:
Where can I download the art of this fix?

OsweRRR:
Se puede hacer un hotfix usando RH_SV_ConnectClient de reapi mirando el código de Karaulov

ghouls96:
> Se puede hacer un hotfix usando RH_SV_ConnectClient de reapi mirando el código de Karaulov

I think so, but how do I get the player's complete set info connection string? get_user_info only allows you to get by keys

I hope soon salt the fix directly in rehlds

Xelson:
> > Se puede hacer un hotfix usando RH_SV_ConnectClient de reapi mirando el código de Karaulov
> 
> I think so, but how do I get the player's complete set info connection string? get_user_info only allows you to get by keys
> 
> I hope soon salt the fix directly in rehlds

`copy_info_string ` from fakemeta

OsweRRR:
> > Se puede hacer un hotfix usando RH_SV_ConnectClient de reapi mirando el código de Karaulov
> 
> I think so, but how do I get the player's complete set info connection string? get_user_info only allows you to get by keys
> 
> I hope soon salt the fix directly in rehlds

https://dev-cs.ru/threads/45970/

dystopm:
@s1lentq took a look on this? seems fine 😁

zik32:
Don't even think about releasing such code. You need to add Q_UnicodeValidate check in several more places.

Nord1cWarr1or:
> Don't even think about releasing such code. You need to add Q_UnicodeValidate check in several more places.

@zik32, what did you mean?

PR #1100, opened 2025-06-23
Fix VectorCompare func while compiling with clang with -flto=full flag enabled

hajimura:
Fixed simd-function VectorCompare which had undefined behavior with -flto=full flag enabled while compiling via Clang
Also maybe this bug could be reproduced with others math fucntions. Need to investigate it further


stamepicmorg:
hello, could you please sync with master please?

PR #1106, opened 2025-07-14
Half-Life 25th Anniversary Update support

stamepicmorg:
- [x] added steam_api libs to windows and linux builds

✅ Windows Ready
✅ Linux Ready since https://github.com/rehlds/ReHLDS/commit/ae2639d3df2478d81836d74ce2342c5f64d4e78b

------------------

Tested on
```
Protocol version 48
Exe version 1.1.2.2 (valve)
Exe build: 19:00:16 Oct  7 2024 (10210
```

1. Installed via 
```
login anonymous
app_set_config 90 mod valve
app_update 90 validate
```
image

2. Installed latest release of rehlds and got error:
image

3. installed `steam_api.dll` from blob from `rehlds` repo.
```
Protocol version 48
Exe version 1.1.2.2/Stdio (valve)
ReHLDS version: 3.14.0.857-dev
Build date: 19:53:00 Mar 27 2025 (4002)
Build from: https://github.com/rehlds/ReHLDS/commit/89958d3
```
image

-------------

linux support:

original
image

rehlds without `libsteam_api.so`
image

rehlds with replaced `libsteam_api.so`

image

-------------
for historical purposes:
https://github.com/rehlds/ReHLDS/actions/runs/16278062208?pr=1106
[win32.zip](https://github.com/user-attachments/files/21222568/win32.zip)
[linux32.zip](https://github.com/user-attachments/files/21222570/linux32.zip)

-------------


stamepicmorg:
oh, we also need to fix windows-builder: https://github.com/actions/runner-images/issues/12045

anzz1:
this rehlds is originally decompiled from hlds 6153 build which is close to 8684, while the new builds are vastly changed with steam networking stuff and such.
you should not bring support to new hlds engine while still using the legacy codebase as base, because it would be mixing versions and could create unforeseen consequences. its better this way that you are required to use the steam_legacy branch aka 8684 so the other libraries match the hlds build ReHLDS is based on.
same is with ReGameDLL_CS, its based on 6153 build i.e. the legacy branch too.

if wanted to move to anniversary version you'd want to decompile the new engine from scratch, both hlds engine and the cs gamedll, and not add any hack to make it work with anniversary (post-8684) libraries.
i dont see any pressing reason to move into the new engine as the 8684 is solid and works just fine, and even on valve's side they have only so far updated halflife, cs, cs:cz to the anniversary while dod, tfc, opfor, dmc, ricochet are still not updated.
maybe after all the mods are updated and valve is done with the anniversary patches a new decompilation in a new anniversary branch could make sense, but it doesn't now.

anzz1:
also i don't understand why are you editing the file information of steam_api.dll in the windows CI, and then signing it with some ReHLDS signature when its an original steam api DLL from valve and is already signed as such, its nonsensical.
its the correct version of 1.60.88.17 that was shipped with hlds 6152/6153 and still is with the 8684 build i.e. legacy branch, so it need not be touched.

anzz1:
> End user should have ability to select which version it will be used - old one from legacy branch or new.

Sure, this could be a goal for the future - but it should be done correctly as in decompile/reverse-engineer the new engine version from scratch - not add a hack to old engine version to mix it with the new one.

> based on old versions of the engine

Exactly. The whole project and rehlds ecosystem incl. regamedllcs is based on the legacy branch. You should not mix versions, that will just invite trouble. For the new engine, a new branch and new reverse-engineering is warranted - not dirty hacks. The whole point of the reverse-engineering project is to match the original HLDS, except when `REHLDS_FIXES` is defined then its fixes are implemented.
A new branch and new reverse-engineering is needed for a new version, not making it a mess of code from different versions messing up the whole clean concept of the project. 

> And in the case of publishing steam_api.dll to understand which versions were installed, we will sign it and edit the metainformation. Probably, slightly changing its appearance, taking into account your comments.

This sentence doesn't make any sense. steam_api.dll is the original file from valve and there is absolutely no need or reason to "sign it and edit the metainformation". It already is signed, and has the version information and such "metainformation". You clearly don't understand what you're doing here, as it doesn't make any sense.

stamepicmorg:
> Sure, this could be a goal for the future - but it should be done correctly as in decompile/reverse-engineer the new engine version from scratch - not add a hack to old engine version to mix it with the new one.

I will take your arguments into consideration and consider them.

> Exactly. The whole project and rehlds ecosystem incl. regamedllcs is based on the legacy branch. You should not mix versions, that will just invite trouble. For the new engine, a new branch and new reverse-engineering is warranted - not dirty hacks. The whole point of the reverse-engineering project is to match the original HLDS, except when `REHLDS_FIXES` is defined then its fixes are implemented.
> A new branch and new reverse-engineering is needed for a new version, not making it a mess of code from different versions messing up the whole clean concept of the project.

So far I have not found any arguments regarding engine compatibility other than the fact that you simply do not like it.
I have re-examined the public changelogs and have not found a single argument that contradicts the addition of support.

Everything that was done here as a "dirty hack" is simply replacing the library with an older version. To restore full backward compatibility.

> This sentence doesn't make any sense. steam_api.dll is the original file from valve and there is absolutely no need or reason to "sign it and edit the metainformation". It already is signed, and has the version information and such "metainformation".

No code hacks were made for support. On the contrary, this compatibility option is currently the most compromise.

As an alternative option, updating the code base for communication with newer versions of the library is being considered.

The fact that you did not like it or did not understand why it was done is your personal business.

I would like to note that this pull request is still open and has not been merged. And work with testing is still underway with it.

If you have any technical arguments, feel free to state them. We will think about it.

> You clearly don't understand what you're doing here, as it doesn't make any sense.

But I will also remind you of the recommendations for adhering to the ethics of communication.
With respect.

PR #1121, opened 2025-08-29
libsteam_api.so forward compat: finish VERSION_SAFE_STEAM_API_INTERFACES implementation

stevenlafl:
## Purpose

This brings InitSafe compatibility with the libsteam_api.so. This should help future-proof as newer libs do not contain symbols for Init().

## Approach

Before this patch, VERSION_SAFE_STEAM_API_INTERFACES results in compile time errors and doesn't run.

This adds ifdef checks where Init() is used, replacing them with InitSafe() as [was already done in other files](https://github.com/rehlds/ReHLDS/blob/5f31306da5842b62736f1be16c0affc2a3c0c27f/rehlds/public/steam/steam_gameserver.h#L49). It properly uses, according to the comments, the CSteamAPIContext as stated in [rehlds/public/steam/steam_api.h](https://github.com/rehlds/ReHLDS/blob/5f31306da5842b62736f1be16c0affc2a3c0c27f/rehlds/public/steam/steam_api.h#L98)

- [ ] Additional longer-term playtesting

## Learning
I explored the headers to gain all this information and kept compiling until it worked. I used gdb to runtime errors. At first, I had compile-time type safety errors, which is how I ended up with the (void *) approach in sv_steam3.cpp, which works flawlessly.

### DIFF
--- a/rehlds/engine/sv_steam3.cpp
+++ b/rehlds/engine/sv_steam3.cpp
@@ -297,9 +297,13 @@ void CSteam3Server::Activate()
 			}
 		}
 
+		#ifdef VERSION_SAFE_STEAM_API_INTERFACES
+		if (!CRehldsPlatformHolder::get()->SteamGameServer_InitSafe(unIP, usSteamPort, gamePort, 0xFFFFu, eSMode, gpszVersionString))
+			Sys_Error("Unable to initialize Steam.");
+		#else
 		if (!CRehldsPlatformHolder::get()->SteamGameServer_Init(unIP, usSteamPort, gamePort, 0xFFFFu, eSMode, gpszVersionString))
 			Sys_Error("Unable to initialize Steam.");
-
+		#endif
 		CRehldsPlatformHolder::get()->SteamGameServer()->SetProduct(gpszProductString);
 		CRehldsPlatformHolder::get()->SteamGameServer()->SetModDir(gamedir);
 		CRehldsPlatformHolder::get()->SteamGameServer()->SetDedicatedServer(g_pcls.state == ca_dedicated);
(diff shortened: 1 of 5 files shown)


anzz1:
And like pointed out in the discussion in that issue #1106 , it's not correct way to proceed to create some from of frankenstein hybrid of the new and old branches.
Both the ReHLDS and its supporting projects like ReGameDLL_CS are reverse-engineered from the old engine the 'legacy' branch. 
And the binary is supposed to be matching to the original as you can see all the changes and fixes are #ifdef'd in a `REHLDS_FIXES` define.
You are simply not supposed to use a different version of steam_api than what is originally designed for. The function signatures change between versions and can lead to undefined behaviour and crashes.
If you look at how steam distribute its games the steam_api DLL is not managed by steam but its distributed part of games files. As in whatever was used for the game when it was compiled that exact version of the API will be shipped and nothing else.
Because sometimes changes are not severe you can indeed get away with using wrong version and it would work just fine until it calls some function that is now in different place in the vtable, which might be never
Its still wrong behaviour and should not be encouraged. It is a *feature* that ReHLDS doesn't start with a wrong API version it is not designed for.

tl;dr; the steam_api.dll and libsteam_api.so must be changed in the CI so they're distributed along with ReHLDS so the correct version is always used

stevenlafl:
> tl;dr; the steam_api.dll and libsteam_api.so must be changed in the CI so they're distributed along with ReHLDS so the correct version is always used

I want to clarify the purpose of this: it is to get the InitSafe() implementation working. That's all this does - whether or not it is packaged with legacy or anniversary wasn't the goal. The anniversary libsteam_api.so does not contain symbols for Init(), only InitSafe().

This is merely a patch to allow it to compile with an updated libsteam_api.so if one so chooses via the VERSION_SAFE_STEAM_API_INTERFACES define. It does not change behavior if undefined.

I should also clarify: this has the **side effect** of allowing one to **overwrite** the anniversary edition with ReHLDS.

PR #1124, opened 2025-09-21
Add CVar: sv_rehlds_custom_gd

anzz1:
#### Add CVar: sv_rehlds_custom_gd <1|0>

0: disallow players joining the server with a different game (default)  
1: allow players joining the server with a different game (e.g. czero -> cstrike)


SmilexGamer:
Unfortunately, we still have the glaring issue of steam authentication not working when connecting from another game and also the fact that the server doesn't appear in server queries.

SmilexGamer:
> Unfortunately, we still have the glaring issue of steam authentication not working when connecting from another game and also the fact that the server doesn't appear in server queries.

Any ideas about this? @anzz1

ScriptedSnark:
> Any ideas about this?

There is already some kind of [solution](https://github.com/jonatan1024/CrossAuth).

SmilexGamer:
> > Any ideas about this?
> 
> There is already some kind of [solution](https://github.com/jonatan1024/CrossAuth).

I'm aware of that repository. However, it doesn't fully fix the problem, it only delays it. The client eventually is kicked from the server.

PR #1125, opened 2025-09-24
fix(userinfo): prohibit ASCII control characters in keys and values

0Ky:
## Description
This PR introduces validation in [`Info_SetValueForStarKey()`](https://github.com/rehlds/ReHLDS/blob/58da97789f3c924dc703bd590f10f5c4a3644543/rehlds/engine/info.cpp#L556-L648) to reject user info keys and values containing ASCII control characters (such as newlines), addressing crashes triggered by malformed user info strings.

## Problem
A malicious actor was detected sending a user info string with a model name containing `\r` (carriage return) or `\n` (line feed) characters, which triggers assertion failures in Windows clients. This led to clients aborting the game or repeatedly hitting the assertion, making ignoring an unacceptable choice.

## Changes
* Adds a check to reject ASCII control characters in both keys and values of user info.

### DIFF
--- a/rehlds/engine/info.cpp
+++ b/rehlds/engine/info.cpp
@@ -589,7 +589,7 @@ qboolean Info_SetValueForStarKey(char *s, const char *key, const char *value, si
 
 	for (const char *p = key; *p; p++)
 	{
-		if (iscntrl(*p))
+		if (iscntrl((unsigned char)*p))
 		{
 			Con_Printf("Can't use keys with ASCII control characters\n");
 			return FALSE;
@@ -598,7 +598,7 @@ qboolean Info_SetValueForStarKey(char *s, const char *key, const char *value, si
 
 	for (const char *p = value; *p; p++)
 	{
-		if (iscntrl(*p))
+		if (iscntrl((unsigned char)*p))
 		{
 			Con_Printf("Can't use values with ASCII control characters\n");
 			return FALSE;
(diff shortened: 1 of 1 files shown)


rehlds/engine/info.cpp
```diff
@@ -587,6 +587,24 @@ qboolean Info_SetValueForStarKey(char *s, const char *key, const char *value, si
 		return FALSE;
 	}
 
+	for (const char *p = key; *p; p++)
+	{
+		if (iscntrl(*p))
```
s1lentq, line 592: please cast to unsigned char
passing a signed char with a negative value to `iscntrl` is UB (undefined behavior)

s1lentq: also add this fix in this two places

https://github.com/rehlds/ReHLDS/blob/94802d2dc42251eb67a05f097bc5e5e4cbd345ca/rehlds/engine/info.cpp#L901
https://github.com/rehlds/ReHLDS/blob/94802d2dc42251eb67a05f097bc5e5e4cbd345ca/rehlds/engine/info.cpp#L924

```diff
		// values should be ended by eos or slash
		while (*s != '\\' && *s != '\0')
		{
			// quotes are deprecated
			if (*s == '"')
				return FALSE;

			// ".." deprecated
			if (*s == '.' && *(s + 1) == '.')
				return FALSE;

+			// control characters are prohibited
+			if (iscntrl((unsigned char)*s))
+				return FALSE;

			s++;
		}
```

PR #1146, opened 2026-01-25
fix(hltv): defer intermission until server is running

### LINKED ISSUE #1115: [BUG] : hltv dies when recording
hltv records every map with at least 2 players on our server and it happens quite often, that while recording the hltv-process dies with:
```
***** FATAL ERROR *****
Server::SetState: not valid m_ServerState (6 -> 8).
*** STOPPING SYSTEM ***
Disconnected.
Master module shutdown.
Status module shutdown.
Demo module shutdown.
World module shutdown.
Server module shutdown.
Network module shutdown.
Proxy module shutdown.
Console log file closed.
```

```
ReHLDS version: 3.15.0.874-dev
ReGameDLL version: 5.28.0.756-dev
Metamod-r v1.3.0.172, API (5:13)
```

this happens since ages, so dont think that any 'recent changes' caused it.

hun1er:
## Purpose
Fixes a crash caused by an invalid state transition when `svc_intermission` arrives before `SERVER_RUNNING`.
Closes #1115

## Approach
Defers intermission handling until after signon completes: if intermission arrives early, set a pending flag and apply `ParseIntermission()` right after the server enters `SERVER_RUNNING`.

### DIFF
--- a/rehlds/HLTV/Core/src/Server.cpp
+++ b/rehlds/HLTV/Core/src/Server.cpp
@@ -121,6 +121,10 @@ bool Server::Init(IBaseSystem *system, int serial, char *name)
 	m_IsVoiceBlocking = false;
 	m_ServerSocket = nullptr;
 
+#ifdef HLTV_FIXES
+	m_PendingIntermission = false;
+#endif
+
 	m_ServerChannel.Create(system);
 	m_ServerInfo.SetMaxSize(512);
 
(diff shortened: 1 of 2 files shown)


s1lentq (approved):

fred0r:
the hltv doesnt seem to die that often than before, but when using [this](https://forums.alliedmods.net/showthread.php?t=109583) amxx plugin, every map - even when theres only the hltv connected - is recorded/created. 
the plugin is configured (since a few years) to only create record, when there are at least two 'real' players and worked fine before.

fred0r:
had `delay 0` set in hltv+amxx  - now trying with `delay 30.0` / `autohltv_delay 30.0`.

PR #1152, opened 2026-02-09
Fix long-standing bug that makes voice mute affect multiple players

anzz1:
## Purpose
Fixes long-standing bug that makes client's voice mute affect multiple players.
https://github.com/ValveSoftware/halflife/issues/557

## Problem
The code used for voice muting is *very* old, dating back to the WON times. The hash of the client's CD key is used to keep a database of players that are muted by the user. This database is held on the client-side (voice_ban.dt).
After transition from WON to Steam, the cd keys weren't used anymore, but the mechanism remained (client->hashedcdkey), instead of taking a hash of the cdkey, now hash of the IP address is taken instead. The client takes hash of its *local* IP address, which wasn't a problem in the pre-NAT era, as people used to be connected straight to the internet and your local IP address was the same as your external IP address.
However, with NAT this mechanism breaks down. The user sends its local IP address as the cdkey, which is always something like 192.168.1.1 or 10.0.0.1, and many users share this local IP. Their "cdkey" hash ends up being the same, so now when you mute a player it will mute multiple players at the same time.

## Solution
To make the mechanism work like pre-NAT times, simply replace the hashedcdkey with the hash of the external IP address of the client. This will make the voice mute mechanism work again like it used to and clients have unique "cdkey" hashes.

### DIFF
--- a/rehlds/engine/sv_main.cpp
+++ b/rehlds/engine/sv_main.cpp
@@ -2500,15 +2500,19 @@ void EXT_FUNC SV_ConnectClient_internal(void)
 	{
 		Con_DPrintf("Client %s connected\nAdr: %s\n", name, NET_AdrToString(host_client->netchan.remote_address));
 	}
-#ifndef REHLDS_OPT_PEDANTIC
+#if !defined(REHLDS_FIXES) && !defined(REHLDS_OPT_PEDANTIC)
 	Q_strncpy(host_client->hashedcdkey, cdkey, 32);
 	host_client->hashedcdkey[32] = '\0';
 #else
 	MD5Context_t ctx;
 	MD5Init(&ctx);
+#ifdef REHLDS_FIXES
+	MD5Update(&ctx, (unsigned char *)&host_client->network_userid.clientip, sizeof(unsigned int));
+#else
 	MD5Update(&ctx, (unsigned char *)cdkey, sizeof(cdkey));
+#endif // REHLDS_FIXES
 	MD5Final((unsigned char *)host_client->hashedcdkey, &ctx);
-#endif
+#endif // !defined(REHLDS_FIXES) && !defined(REHLDS_OPT_PEDANTIC)
 
 	host_client->active = FALSE;
 	host_client->spawned = FALSE;
@@ -4047,7 +4051,7 @@ void EXT_FUNC SV_WriteFullClientUpdate_internal(IGameClient *client, char *info,
 {
 	client_t* cl = client->GetClient();
 
-#ifndef REHLDS_OPT_PEDANTIC
+#if !defined(REHLDS_FIXES) && !defined(REHLDS_OPT_PEDANTIC)
 	unsigned char digest[16];
 
 	MD5Context_t ctx;
@@ -4061,7 +4065,7 @@ void EXT_FUNC SV_WriteFullClientUpdate_internal(IGameClient *client, char *info,
 	MSG_WriteLong(sb, cl->userid);
 	MSG_WriteString(sb, info);
 
-#ifndef REHLDS_OPT_PEDANTIC
+#if !defined(REHLDS_FIXES) && !defined(REHLDS_OPT_PEDANTIC)
 	MSG_WriteBuf(sb, sizeof(digest), digest);
 #else
 	MSG_WriteBuf(sb, 16, cl->hashedcdkey);


Th3-822:
i can't believe how this was even merged without being on a sv_lan 1 check or doing it in a safer way...
surely there are ppl waiting for servers to update to this version to start DDoSing admins/players ips thanks to this ~~fix~~ security issue

anzz1:
@Th3-822 You are fundamentally misunderstanding the situation. This doesn't reveal ips, they are hashed. Not that revealing ip is much of a security issue in this day and age. Everywhere you ever connect knows your ip. (and yes i know hashes can be cracked but again it is not much of an issue. you could add random salt by server to the hash which would make it uncrackable but then youd lose the intended feature of having voice ban carry over servers).

Th3-822:
Exposing all players ips on a public online server is kinda of a really bad idea unless they are all using VPN
and about the reliable method to carry the voice ban" over servers. Well, there's *sid/authid... ~~(as long as there is nothing installed that compromises that)~~

anzz1:
it's not "kinda of a really bad idea" lot of major games today (GTAV and COD franchise come to mind for some "small" franchises) that operate entirely on p2p principle and everyone know everyone's ip and its not a problem at all. ip should be considered public on the internet and routers drop traffic that don't belong to you anyway. and there is even added layer of security that not ips are known but hashes of them. you are worrying over nothing.

that being said it could also use a hash of the authid, or alternatively if the ip is salted as following adding a 128-bit key to the 32-bit ip, the resulting hashed value will be 160-bit which makes it completely uncrackable

```diff
diff --git a/rehlds/engine/sv_main.cpp b/rehlds/engine/sv_main.cpp
index 38943b7..3282262 100644
--- a/rehlds/engine/sv_main.cpp
+++ b/rehlds/engine/sv_main.cpp
@@ -2508,6 +2508,7 @@ void EXT_FUNC SV_ConnectClient_internal(void)
 	MD5Init(&ctx);
 #ifdef REHLDS_FIXES
 	MD5Update(&ctx, (unsigned char *)&host_client->network_userid.clientip, sizeof(unsigned int));
+	MD5Update(&ctx, (unsigned char *)g_ServerKeyStore.GetServerKey(), 16);
 #else
 	MD5Update(&ctx, (unsigned char *)cdkey, sizeof(cdkey));
 #endif // REHLDS_FIXES
diff --git a/rehlds/rehlds/rehlds_security.cpp b/rehlds/rehlds/rehlds_security.cpp
index a8edbb0..1b9e0f2 100644
--- a/rehlds/rehlds/rehlds_security.cpp
+++ b/rehlds/rehlds/rehlds_security.cpp
@@ -12,6 +12,7 @@ cvar_t sv_rehlds_stringcmdrate_burst_punish = { "sv_rehlds_stringcmdrate_burst_p
 
 CMoveCommandRateLimiter g_MoveCommandRateLimiter;
 CStringCommandsRateLimiter g_StringCommandsRateLimiter;
+CServerKeyStore g_ServerKeyStore;
 
 CMoveCommandRateLimiter::CMoveCommandRateLimiter() {
 	Q_memset(m_AverageMoveCmdRate, 0, sizeof(m_AverageMoveCmdRate));
@@ -175,6 +176,35 @@ void CStringCommandsRateLimiter::CheckAverageRate(unsigned int clientId) {
 	}
 }
 
+CServerKeyStore::CServerKeyStore() {
+	Q_memset(m_ServerKey, 0, sizeof(m_ServerKey));
+}
+
+const unsigned char* CServerKeyStore::GetServerKey() {
+	if (m_ServerKey[0] == 0) {
+		size_t n;
+		FILE* f = fopen("server.key", "rb");
+		if (f) {
+			n = fread(m_ServerKey, 16, 1, f);
+			fclose(f);
+		}
+		if (n != 1 || m_ServerKey[0] == 0) {
+			std::random_device rd;
+			for (int i = 0; i < 4; ++i) {
+				uint32_t r = rd();
+				memcpy(m_ServerKey + i * 4, &r, 4);
+			}
+			if (m_ServerKey[0] == 0) m_ServerKey[0]++;
+			f = fopen("server.key", "wb");
+			if (f) {
+				fwrite(m_ServerKey, 16, 1, f);
+				fclose(f);
+			}
+		}
+	}
+	return m_ServerKey;
+}
+
 void Rehlds_Security_Init() {
 #ifdef REHLDS_FIXES
 	Cvar_RegisterVariable(&sv_rehlds_movecmdrate_max_avg);
diff --git a/rehlds/rehlds/rehlds_security.h b/rehlds/rehlds/rehlds_security.h
index c287314..f1106de 100644
--- a/rehlds/rehlds/rehlds_security.h
+++ b/rehlds/rehlds/rehlds_security.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "engine.h"
+#include <random>
 
 class CMoveCommandRateLimiter {
 public:
@@ -42,6 +43,17 @@ private:
 
 extern CStringCommandsRateLimiter g_StringCommandsRateLimiter;
 
+class CServerKeyStore {
+public:
+	CServerKeyStore();
+	const unsigned char* GetServerKey();
+
+private:
+	unsigned char m_ServerKey[16];
+};
+
+extern CServerKeyStore g_ServerKeyStore;
+
 extern void Rehlds_Security_Init();
 extern void Rehlds_Security_Shutdown();
 extern void Rehlds_Security_Frame();
```

but that like the authid would veer away from the original functionality. the old code use hash of ip after all and the new code makes no difference to anyone who is directly corrected to internet, so it was always "exposed" before too, the NAT is what obscured it.

PR #1157, opened 2026-03-01
Fix "executable stack" errors

### LINKED ISSUE #1079: [BUG]: engine_i486.so: cannot enable executable stack as shared object requires: Invalid argument Unable to load engine, image is corrupt.
### 👉👈 Contact Details (optional)

https://steamcommunity.com/id/_SmileY

### 🤔 What happened?

An error while running ReHLDS on arch linux

```
[eu@pc 27020]$ ./start.sh 
Enabling debug mode
./hlds_run: linha 125: test: unlimited: esperava expressão de número inteiro
gdb: warning: Couldn't determine a path for the index cache directory.
Error: /home/eu/Desktop/27020/engine_i486.so: cannot enable executable stack as shared object requires: Invalid argument
Unable to load engine, image is corrupt.
cat: hlds.9982.pid: Arquivo ou diretório inexistente
                                                    email debug.log to linux@valvesoftware.com
                                                                                              sex 07 fev 2025 22:12:32 -03: Server Quit
```

Something related with engine_i486.so.

An workaround is call execstack -c engine_i486.so and now ReHLDS works again. But is not the ideal scenario right?

Anyway i hope someone can fix it

### ⚠️ Meta-information

```shell
- No plugins, server can´t run
- No output, server can´t run
   -> Latest ReHLDS master build: https://github.com/rehlds/ReHLDS/actions/runs/12501656147
   -> Latest ReGameDLL_CS master build: https://github.com/rehlds/ReGameDLL_CS/actions/runs/12333552178
- No Plugins!
- Running Arch Linux
```

### 📄 Relevant log output

```shell
[eu@pc ~]$ fastfetch 
                  -`                     eu@pc
                 .o+`                    -----
                `ooo/                    OS: Arch Linux x86_64
               `+oooo:                   Kernel: Linux 6.13.1-zen1-1-zen
              `+oooooo:                  Uptime: 18 mins
              -+oooooo+:                 Packages: 739 (pacman)
            `/:-:++oooo+:                Shell: bash 5.2.37
           `/++++/+++++++:               Display (Tempest  34'): 3440x1440 @ 144 Hz in 34" [External]
          `/++++++++++++++:              DE: Cinnamon 6.4.6
         `/+++ooooooooooooo/`            WM: Muffin (X11)
        ./ooosssso++osssssso+`           WM Theme: Fluent-Dark (Adwaita)
       .oossssso-`/ossssss+`          Theme: Fluent-Dark [GTK2/3/4]
      -osssssso.      :ssssssso.         Icons: Fluent-dark [GTK2/3/4]
     :osssssss/        osssso+++.        Font: Ubuntu (10pt) [GTK2/3/4]
    /ossssssss/        +ssssooo/-        Cursor: Fluent-dark (24px)
  `/ossssso+/:-        -:/+osssso+-      Terminal: GNOME Terminal 3.54.3
 `+sso+:-`                 `.-/+oso:     Terminal Font: Source Code Pro (10pt)
`++:.                           `-/+/    CPU: Intel(R) Core(TM) i5-7400 (4) @ 3.50 GHz
.`                                 `/    GPU: NVIDIA GeForce GTX 1050 Ti [Discrete]
                                         Memory: 2.83 GiB / 15.56 GiB (18%)
                                         Swap: 0 B / 4.00 GiB (0%)
                                         Disk (/): 26.98 GiB / 438.06 GiB (6%) - ext4
                                         Local IP (enp6s0): 192.168.100.100/24
                                         Locale: pt_BR.UTF-8

                                                                 
                                                                 
[eu@pc ~]$
```

mxpph:
## Purpose
Fixes #1079 

## Approach
Adds the compiler and linker flags `-Wa,--execstack` and `-Wl,-z,execstack` respectively, as proposed by @anzz1 in the discussion on that issue.

I have tested these fixes and no longer crash due to the "executable stack" error on both Debian 13 and Fedora 42 (both with glibc 2.41).

### DIFF
--- a/rehlds/HLTV/Console/CMakeLists.txt
+++ b/rehlds/HLTV/Console/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 3.1)
 project(hltv CXX)
 
+include(CheckCXXCompilerFlag)
+
 option(DEBUG "Build with debug information." OFF)
 
 set(CMAKE_CXX_STANDARD 14)
@@ -13,6 +15,18 @@ set(COMPILE_FLAGS "-m32 -fno-pic -fno-pie -U_FORTIFY_SOURCE -DFORTIFY_SOURCE=0")
 set(LINK_FLAGS "-m32 -no-pie -Wl,--no-export-dynamic")
 
 set(COMPILE_FLAGS "${COMPILE_FLAGS} -Wall -fno-exceptions -fno-strict-aliasing -fno-strict-overflow")
+
+check_cxx_compiler_flag("-Wa,--execstack" HAS_EXECSTACK_ASM)
+check_cxx_compiler_flag("-Wl,-z,execstack" HAS_EXECSTACK_LINK)
+
+# Glibc >=2.41 fixes
+if(HAS_EXECSTACK_ASM)
+	set(COMPILE_FLAGS "${COMPILE_FLAGS} -Wa,--execstack")
+endif()
+
+if(HAS_EXECSTACK_LINK)
+	set(LINK_FLAGS "${LINK_FLAGS} -Wl,-z,execstack")
+endif()
 
 if (DEBUG)
 	set(COMPILE_FLAGS "${COMPILE_FLAGS} -g3 -O3 -ggdb")
--- a/rehlds/dedicated/CMakeLists.txt
+++ b/rehlds/dedicated/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 3.1)
 project(hlds CXX)
 
+include(CheckCXXCompilerFlag)
+
 option(DEBUG "Build with debug information." OFF)
 
 set(CMAKE_CXX_STANDARD 14)
@@ -13,6 +15,18 @@ set(COMPILE_FLAGS "-m32 -fno-pic -fno-pie -U_FORTIFY_SOURCE -DFORTIFY_SOURCE=0")
 set(LINK_FLAGS "-m32 -no-pie -Wl,--no-export-dynamic")
 
 set(COMPILE_FLAGS "${COMPILE_FLAGS} -Wall -fno-exceptions -fno-strict-aliasing -fno-strict-overflow")
+
+check_cxx_compiler_flag("-Wa,--execstack" HAS_EXECSTACK_ASM)
+check_cxx_compiler_flag("-Wl,-z,execstack" HAS_EXECSTACK_LINK)
+
+# Glibc >=2.41 fixes
+if(HAS_EXECSTACK_ASM)
+	set(COMPILE_FLAGS "${COMPILE_FLAGS} -Wa,--execstack")
+endif()
+
+if(HAS_EXECSTACK_LINK)
+	set(LINK_FLAGS "${LINK_FLAGS} -Wl,-z,execstack")
+endif()
 
 if (DEBUG)
 	set(COMPILE_FLAGS "${COMPILE_FLAGS} -g3 -O3 -ggdb")


anzz1:
Not including the steam_api.dll and libsteam_api.so, so they have to come from the *legacy* branch when downloading hlds is actually a good feature. That if the user doesn't realize to change to the *legacy* branch, they will crash with "procedure entry SteamApps not found", "undefined symbol: SteamGameServer_Init" or other such error. It is a good thing, as if you include the steam_api and libsteam_api in the rehlds archive, the user might accidentally download the *anniversary* branch and mix two engine version together, which is a **big no-no**.

I've explained the problem in more detail in these two comments:
https://github.com/rehlds/ReHLDS/issues/1149#issuecomment-3873358145
https://github.com/rehlds/ReHLDS/issues/1149#issuecomment-3880284096

Otherwise i don't see a problem, except a similar gcc version check should be added like you can already see with the `--fcf-protection`, that you only add the `execstack` to a gcc version that understands it
```cmake
# GCC >= 8.3
if (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 8.0)
	set(COMPILE_FLAGS "${COMPILE_FLAGS} -fcf-protection=none")
endif()
```
Or is it supported by older GCC versions too, so such guard isn't needed? I've not done the research just asking whether you've confirmed this.

anzz1:
Something like this should probably work
```cmake
include(CheckCXXCompilerFlag)

# Glibc >=2.41 fix
check_cxx_compiler_flag("-Wa,--noexecstack" HAS_NOEXECSTACK_ASM)
check_cxx_compiler_flag("-Wl,-z,noexecstack" HAS_NOEXECSTACK_LINK)
if(HAS_NOEXECSTACK_ASM)
	set(COMPILE_FLAGS "${COMPILE_FLAGS} -Wa,--noexecstack")
endif()
if(HAS_NOEXECSTACK_LINK)
	set(LINK_FLAGS "${LINK_FLAGS} -Wl,-z,noexecstack")
endif()
```

I already checked that this CheckCXXCompilerFlag function exist in 3.1 cmake which is the current minimum version, so it should work without further changes. i didn't test to compile though.

mxpph:
> Not including the steam_api.dll and libsteam_api.so, so they have to come from the _legacy_ branch when downloading hlds is actually a good feature. That if the user doesn't realize to change to the _legacy_ branch, they will crash with "procedure entry SteamApps not found", "undefined symbol: SteamGameServer_Init" or other such error. It is a good thing, as if you include the steam_api and libsteam_api in the rehlds archive, the user might accidentally download the _anniversary_ branch and mix two engine version together, which is a **big no-no**.

Understood. I'll incorporate your suggestions.

mxpph:
Thanks for your snippet, I tested and it builds and runs just fine.

anzz1:
Adding to this, my understanding is that the problem comes from trying to load shared objects with an executable stack while the main executable doesn't have it set. Meaning that dynamically loaded libraries (shared objects) cannot change the state of the executable stack, not that they aren't allowed per-se.

Where I'm getting at here is mods and their gamedlls, whatever libraries they might load, plus metamod and its modules, could run into the same issue where they're marked as requiring executable stack and fail to load.

Therefore the option of simply setting the stack executable to retain original functionality and compatibility would probably be the right way forward. That is instead of having the shared objects declared as not having executable stack, declare the main hlds binary having it, therefore allowing all the loaded modules to either have or not have it.

So instead of changing the makefiles for the modules, change it for the `hlds_linux` and `hltv` executables, and instead of `noexecstack` do the opposite: `-Wa,--execstack`and `-Wl,-z,execstack`

I don't have a machine with glibc 2.41 to test, could you test if that works?

mxpph:
It compiles and runs, but worth noting that I get this warning:
```
/usr/bin/ld: warning: cputype32.o: missing .note.GNU-stack section implies executable stack
/usr/bin/ld: NOTE: This behaviour is deprecated and will be removed in a future version of the linker
```
This is because the object files in `rehlds/lib/linux32/libacof32.a` were assembled from raw assembly .S files that don't contain the .note.GNU-stack section. They can be found here. https://github.com/lukego/asmlib/tree/master

Issue #1165, opened 2026-04-06
[BUG]: An exploit allowing to set fake angles whenever the player wants to.


autisoid:
### 👉👈 Contact Details (optional)

kirka@nigge.rs (yes this is a real email)

### 🤔 What happened?

The engine has a strange part of code which effectively allows a hacker to set whatever the viewangles they want to have, meanwhile allowing them to interact with the world normally, that is, shooting, using buttons, etc.
https://github.com/rehlds/ReHLDS/blob/master/rehlds/engine/sv_user.cpp#L1644-L1646
This very part of code sets the angles seen by others players based on the very last movement command sent by the current client, so if a hacker appends a fake command containing the angles they want to have, this code will eat them.
Basically this:
```cpp
/*
=================
CFakeAngles::OnEvent

The shi responsible for setting fake angles.
Works by adding a movement command with the
angles desired.
=================
*/
void CFakeAngles::OnEvent( const ISimpleEvent* _Event )
{
	if ( !g_pClientDesignerModule->m_pEnableFakeAngles->Get( ) )
		return;

	if ( _Event->GetType( ) == EEventType::kCL_MoveEvent ) // CL_Move enter
	{
		m_fInMove = true;
		m_fHasJustEnteredMove = true;
	}

	// x: Trivia: CNetworkUserCommandSubmitEvent has m_pBuffer, m_pTo and m_pFrom members,
	// which are references to the parameters passed to MSG_WriteUsercmd respectively:
	// void MSG_WriteUsercmd( sizebuf_t* buf, usercmd_t* to, usercmd_t* from )
	if ( _Event->GetType( ) == EEventType::kNetworkUserCommandSubmitEvent ) // MSG_WriteUsercmd
	{
		const CNetworkUserCommandSubmitEvent* e = static_cast< const CNetworkUserCommandSubmitEvent* >( _Event );

		// Prevent modifying user calls if any.
		if ( !m_fInMove )
		{
			return;
		}
		// No recursion.
		if ( m_fRecurseGuard )
		{
			return;
		}

		// Calculate the number of commands left to write.
		// Only once, per CL_Move entrance.
		if ( m_fHasJustEnteredMove )
		{
			m_nCmdsLeft = g_iNumBackup + g_nModifiedNumCmds;
			m_fHasJustEnteredMove = false;
		}

		m_nCmdsLeft--;

		m_fRecurseGuard = true;
[...]

PR #1172, opened 2026-05-01
 [ engine ] Add built-in speedhack detection

autisoid:
## Purpose
Adds a built-in speedhack detection.

## Approach
Based on counting the client timer and counting approximate client's base game speed.

#### Open Questions and Pre-Merge TODOs
- [x] Add the convars to readme
- [x] Rewrite CGameClient's stuffs because it looks ugly right now

## Learning
The client can use the inbuilt `sys_timescale` hidden convar (has to be registered first) or Cheat Engine speedhack to speed up or slow down the game.

`sv_rehlds_movement_max_timescale` controls how much the client can speed up the _base game speed_, and `sv_rehlds_movement_min_timescale` controls how much the client can slow it down.

`sv_rehlds_movement_max_error_msec` controls how far ahead client's internal timer can go ahead the server's timer. Going ahead does nothing until client's _base game speed_ is lower/higher than `sv_rehlds_movement_min_timescale`/`sv_rehlds_movement_max_timescale`

Based on: wootguy's anticheat (https://github.com/wootguy/AntiCheat/blob/master/src/anticheat.cpp), LevShisterov's researches.

Resolves https://github.com/rehlds/ReHLDS/issues/1164.

PR #1173, opened 2026-05-19
[ DRAFT ] [ engine ] ( NEEDS TESTING A LOT ) Add sv_rehlds_movecmds_holdaim_backlog

autisoid:
## Purpose
Break fake angles/perfect silent exploits (#1165)

## Approach
The server is now taking several snapshots of viewangles sent by the client, and sets angles viewable by other players based on the very last record saved. This does not fully mitigate the fake angles/perfect silent exploits, as there's a bypass I'm aware of (in theory, although) but currently none of the hacks available to the public are able to bypass this.

- [ ] Perform very deep testing process.

## Learning
The attacker has to either use code from #1165 or make the exploit on their own. Basically, the player has to send two movement commands (not packets!) in the very same frame, i.e. `numcmds == 2` in a single `clc_move`, which can be achieved by setting `cls.nextcmdtime = FLT_MAX;` for a frame. The first frame (i.e. the frame we've set `cls.nextcmdtime` to `FLT_MAX`) has to have angles untouched, and the very next frame has to have angles wanted and `cls.nextcmdtime` has to be reset to `-1.0f`.

Pseudocode:

```cpp
VOID CDECL CL_CreateMove( FLOAT _Frametime, struct usercmd_s* _Command, INT _Active )
{
    // Normal code goes before this. This has to be placed in the very end of the function.
    
    if ( !g_fDoPerfectSilentFlip ) return;

    if ( cls.nextcmdtime == FLT_MAX )
    {
        cls.nextcmdtime = -1.0f; // send the command NOW
        g_fDoPerfectSilentFlip = FALSE; // we did the flip
    }
    else
    {
        _Command->viewangles[ YAW ] -= 180.0f;
        _Command->buttons |= IN_ATTACK; // shoot backwards
        cls.nextcmdtime = FLT_MAX;
    }
}

/*
=================
CL_Flip_f

=================
*/
VOID CDECL CL_Flip_f( VOID )
{
    g_fDoPerfectSilentFlip = TRUE;
}
```

Register `CL_Flip_f` as an in-game command and execute it at any moment. The player will shoot a bullet from their back, but none of the players in the server (including yourself) will be able to see that you did rotate to do so. There will be a decal from the bullet shot.

### DIFF
--- a/rehlds/rehlds/rehlds_security.cpp
+++ b/rehlds/rehlds/rehlds_security.cpp
@@ -10,8 +10,19 @@ cvar_t sv_rehlds_movecmdrate_burst_punish = { "sv_rehlds_movecmdrate_burst_punis
 cvar_t sv_rehlds_stringcmdrate_avg_punish = { "sv_rehlds_stringcmdrate_avg_punish", "5", 0, 5.0f, NULL };
 cvar_t sv_rehlds_stringcmdrate_burst_punish = { "sv_rehlds_stringcmdrate_burst_punish", "5", 0, 5.0f, NULL };
 
+cvar_t sv_rehlds_movecmd_max_ticks = { "sv_rehlds_movecmd_max_ticks", "24", 0, 24.0f, NULL };
+cvar_t sv_rehlds_movecmd_max_null_streak = { "sv_rehlds_movecmd_max_null_streak", "0", 0, 0.0f, NULL };
+cvar_t sv_rehlds_movecmd_clamp_interp = { "sv_rehlds_movecmd_clamp_interp", "1", 0, 1.0f, NULL };
+cvar_t sv_rehlds_movecmdtime_samples = { "sv_rehlds_movecmdtime_samples", "120", 0, 120.0f, NULL };
+cvar_t sv_rehlds_movecmdtime_max_error = { "sv_rehlds_movecmdtime_max_error", "300", 0, 300.0f, NULL };
+cvar_t sv_rehlds_movecmdtime_max_scale = { "sv_rehlds_movecmdtime_max_scale", "3.0", 0, 3.0f, NULL };
+cvar_t sv_rehlds_movecmdtime_min_scale = { "sv_rehlds_movecmdtime_min_scale", "0.5", 0, 0.5f, NULL };
+cvar_t sv_rehlds_movecmdtime_punish = { "sv_rehlds_movecmdtime_punish", "-1", 0, -1.0f, NULL };
+cvar_t sv_rehlds_movecmdtime_max_warnings = { "sv_rehlds_movecmdtime_max_warnings", "-1", 0, -1.0f, NULL };
+
 CMoveCommandRateLimiter g_MoveCommandRateLimiter;
 CStringCommandsRateLimiter g_StringCommandsRateLimiter;
+CUserCmdTimeLimiter g_UserCmdTimeLimiter;
 
 CMoveCommandRateLimiter::CMoveCommandRateLimiter() {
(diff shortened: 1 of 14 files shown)


anzz1:
I think this and #1172 and other such anti-hack measures that add extra processing to the server would be better served as a separate metamod plugin utilizing the rehlds api and its hooks. And if hook for a certain function doesn't exist, rather create the hook and make a PR out of it.

This way extra code that isn't necessarily vanilla engine related but rather an additional plugin, could be put in a plugin which would both make it easy for server owners to choose whether to use it, but also help future maintainability as the code that is the server and the regular engine vs. extra functionality like anti-cheat would be in their separate codebases and repositories.

If you are unsure how to create plugins here is an example of a very simple hook plugin that was used to fix an issue that has since been incorporated to the main rehlds repo: https://github.com/anzz1/revoicemutefix
Making a similar thing called ReAntiCheat or whatever would use the modular approach that is possible with the ReHLDS API and certainly seems like the correct way to do this.

In my opinion b8f7214 should be reverted too and that speedhack detection feature added to the plugin. Such plugin repo could also have more contributors with push access so it could have faster turnaround of development as experimental things could be tried easier than with base ReHLDS that needs to stay rock-solid and diligently tested as everyone depends on it.

Further on, a separate anticheat plugin could also include experimental features that would have high resource usage and never belong in base ReHLDS like raycast based anti-wallhack that'd selectively not send position of players that are behind walls to other players. The more you think about it, the surer I am that anti-cheat features should be in its own separate plugin and it could be packed full of features in the end.

anzz1:
Now that it's closed, are you creating a new repo to make an anti-cheat plugin? I'd like to follow the progress

autisoid:
It actually got closed automatically without me noticing it at all.

anzz1:
Alright, so have you considered reworking it into a plugin? Don't think anti cheat belongs in ReHLDS core.

rehlds/rehlds/rehlds_security.cpp
```diff
@@ -23,6 +23,7 @@ cvar_t sv_rehlds_movecmdtime_max_warnings = { "sv_rehlds_movecmdtime_max_warning
 CMoveCommandRateLimiter g_MoveCommandRateLimiter;
 CStringCommandsRateLimiter g_StringCommandsRateLimiter;
 CUserCmdTimeLimiter g_UserCmdTimeLimiter;
+CUserCmdAnglesBacklog g_CUserCmdAnglesBacklog;
```
count-fahqweh, line 26: Hello! Compilation seems to be wrong because of `g_CUserCmdAnglesBacklog` that's been copy-pasted. That should be `g_UserCmdAnglesBacklog` !

count-fahqweh:
Hello! I've been testing this a little in Half-Life 1 (NOT Counter-Strike 1.6) and noticed that if `sv_rehlds_movecmds_holdaim_backlog` is set below a value of `8`, spectating is completely messed up, resulting in super weird angles. Third person angles will also look very weird.

autisoid:
Appreciating the will to try that out. Yeah I seem to have copy paste that as I've been copying the code from my local branch of rehlds with other custom code. Will fix and look into the problem of values < 8

autisoid:
@anzz1 I think at least the anti speedhack should be a part of the engine and should be enabled out of the box. Because there are too many free hacks with speedhack abilities, as well as external stuff with the speedhack ability just like Cheat Engine and others. If the anti speedhack stuffs are in the engine, we make sure nobody does use (at least) speedhack on latest rehlds-powered servers...
TL;DR: speedhack in my opinion is a type of hack which should be prevented out of the box

Although other thoughts about an anti-wallhack based on raycasting in a plugin are completely OK. I double that.

PR #1176, opened 2026-06-01
fix: add NULL check for host_client->edict in SV_RunCmd

fred0r:
Prevents SIGSEGV when host_client->edict is NULL while dereferencing v.solid at sv_user.cpp:1053. The edict pointer can become NULL during entity callbacks (e.g. pfnCmdStart, pfnPlayerPreThink, pfnPM_Move) when a plugin triggers SV_DropClient within the same frame.

## Purpose

HLDS crashes with SIGSEGV (signal 11) at `engine_i486.so` offset `0x2304b` inside
`SV_RunCmd`. The instruction `mov 0x18c(%ecx),%esi` dereferences `host_client->edict->v.solid`
where `host_client->edict` is NULL. Fault address: `0x18c`.

Trigger: an AMXX plugin (ATAC team-attack counter + AMXBans) calls `SV_DropClient` via
the command buffer during an entity callback that fires inside `SV_RunCmd` before line 1053,
notably `pfnCmdStart`, `pfnPlayerPreThink`, `pfnPlayerThink`, or `pfnPM_Move`. Once
`host_client->edict` is nulled by `SV_DropClient_internal` in `host.cpp:518`, any
subsequent dereference in the same frame crashes.

Reproducible across multiple maps; confirmed in 5+ core dumps and minidumps from
a CS 1.6 server running reHLDS 3.15.0.896 + AMXX 1.10.0.5476.

## Approach

Add a single NULL guard: `if (host_client->edict && host_client->edict->v.solid)`.
When the edict pointer is NULL the entire `SV_LinkEdict` / touch-handling block is
skipped, which is safe because there is no entity to link. The function then proceeds
to `pfnPlayerPostThink` and `pfnCmdEnd`

The fix is minimal — one addition — and does not alter any logic paths for valid entities.

- [x] Verified binary disassembly matches source: `0x18c` = `edict->v.solid` (edict offset `0x80` + entvars `solid` at `0x10c`)
- [x] Confirmed 5 independent crash dumps share the same EIP
- [x] Root cause traced to `SV_DropClient` nulling `cl->edict` mid-frame via plugin callbacks
- [X] Backport to upstream reHLDS: the same bug exists in ReHLDS `master` at `sv_user.cpp:1053`

## Learning

The failed `pClient->edict` pointer at offset `0x4a84` in `client_t` was identified by
tracing the crash EIP to the disassembly, cross-referencing with `progdefs.h` (`entvars_t`
layout) and `server.h` (`client_s` layout). The engine misses the NULL check because
vanilla HLDS never exposes `host_client->edict` to callbacks that can drop the client
— this became visible only with AMXX/metamod hooking the entity interface at every
movement callback stage.

`addr2line` confirmed: `SV_RunCmd(usercmd_s*, int, int, int)`.
`eu-readelf -n` on the core dump provided signal, register, and memory-map data.
`objdump -d` provided the exact instruction-level trace.


justgo97:
A small plugin to reproduce the issue would be useful

Also fixes should be wrapped around REHLDS_FIXES macro.

fred0r:
```
#include <amxmodx>
#include <amxmisc>
#include <reapi>

#define PLUGIN "ReHLDS Crash Test"
#define VERSION "1.3"

new bool:g_test_active

public plugin_init() {
	register_plugin(PLUGIN, VERSION, "fred0r")

	register_concmd("amx_crashtest_reapi", "cmd_reapi", ADMIN_RCON,
		"- trigger crash via rh_drop_client in PreThink")
	register_concmd("amx_crashtest_disarm", "cmd_disarm", ADMIN_RCON,
		"- disarm crash test")
}

public cmd_reapi(id, level, cid) {
	if (!cmd_access(id, level, cid, 1))
		return PLUGIN_HANDLED

	g_test_active = true
	console_print(id, "[Test] ReAPI armed — crash on next client_PreThink (unpatched)")
	return PLUGIN_HANDLED
}

public cmd_disarm(id, level, cid) {
	if (!cmd_access(id, level, cid, 1))
		return PLUGIN_HANDLED

	g_test_active = false
	console_print(id, "[Test] Disarmed")
	return PLUGIN_HANDLED
}

public client_PreThink(id) {
	if (!g_test_active)
		return

	if (!is_user_connected(id))
		return

	g_test_active = false
	rh_drop_client(id, "Crash test")
}

```

fred0r:
@justgo97
can you confirm, pls?

JUNESYNGOTOFLEX:
@fred0r 
i can confirm this crash
i got this from one of the AFK plugins

code tested
```
#include <amxmodx>
#include <reapi>

new gState = false;

public plugin_init() {
	// Any player tnink hook
	RegisterHookChain(RG_CBasePlayer_PreThink, "CBasePlayer_PreThink", 0); // same in post
	
	// Cmd to trigger
	register_clcmd("say /kick", "clcmd_trigger");
}

public clcmd_trigger(id) {
	gState = true;
	return 1;
}

public CBasePlayer_PreThink(const id) {
	if (gState) {
		rh_drop_client(id, "AFK");
		
		// Same, but without ReAPI
		// server_cmd("kick #%d ^"AFK^"", get_user_userid(id));
		// server_exec();
		
		gState = false;
	}
}
```

fred0r:
@JUNESYNGOTOFLEX 
good to hear that - apparently a lot ppl test with bots only and not with 'real' players on server.

JUNESYNGOTOFLEX:
@fred0r 
I've received complaints about this crash from some server owners, but I didn't really understand the reason at the time. I solved it by using this function with a 1-frame delay. Now I see the issue and can confirm that it is relevant.

fred0r:
sry, but what will happen now with this pr? 
any feedback would be nice

fred0r:
https://github.com/rehlds/ReHLDS/pull/1199

Issue #1098, opened 2025-06-09
[BUG]: new rehlds Exploit (LAG)???


darth19:
### 🤔 What happened?

hacker is Lagging my server, and this appear in the console,  (everyone hit 1000 ping)
im using the latest rehlds
Exe version 1.1.2.7/Stdio (cstrike)
         ReHLDS version: 3.14.0.857-dev
         Build date: 19:52:21 Mar 27 2025 (4002)
         Build from: https://github.com/rehlds/ReHLDS/commit/89958d3
[https://i.ibb.co/x8ByGs4s/votan-server-junio-2025-20.png](url)

https://i.ibb.co/x8ByGs4s/votan-server-junio-2025-20.png

### ⚠️ Meta-information

```shell
ReHLDS version: 3.14.0.857-dev
         Build date: 19:52:21 Mar 27 2025 (4002)
         Build from: https://github.com/rehlds/ReHLDS/commit/89958d3
linux ubuntu
```

### 📄 Relevant log output

```shell

```

Issue #1054, opened 2024-10-05
New Exploit ReHLDS ?


wilianmaique:
I'm receiving an 'attack', but it's not a DDoS. As soon as the attack happens, I quickly restart the map, and the attack disappears instantly, which confirms it's some kind of 'exploit', but I still don't have enough details.

Note: It's a '4fun' server with not many plugins, and none of the plugins increase the ping, etc.

I'm posting here to see if anyone has more details.

HypeGfx:
Yeah this is a new exploit I experienced it too recently...

wilianmaique:
> Yeah this is a new exploit I experienced it too recently...
Do you know any way to block etc?

SmilexGamer:
Logs would be helpful.

wilianmaique:
> [SmilexGamer](/SmilexGamer)

I don't have logs, it's an exploit, hard to know

SmilexGamer:
> > [SmilexGamer](/SmilexGamer)
> 
> I don't have logs, it's an exploit, hard to know

With no logs, there's not much the devs can do to fix this.

HypeGfx:
Ok I'm not certain if this will fix it but try to check your mp_consistency and set it to 1.. I'm just changing a lot of ConVars desperate to find a reason for it.. I will reply on this topic in like 12-15 hr...

rishhh78:
Let me explain the issue... 

Basically what happens is, player does something I have no idea but it happens when is in the server (It's most likely related to map thingy i guess) so server resource gets utilized maximum and players starts to lag badly as if its like some DDoS attack.... server memory ramps up in rapid amount as well.. 

I got one report here not sure if its related to that one [https://hernan.de/blog/lock-and-load-exploiting-counter-strike-via-bsp-map-files/](url)

di57inct:
try setting sv_send_logos 0 and sv_allowupload 0 and if you're using fast download sv_allowdownload 0. there's also sv_allow_dlfile too but idk what that does exactly. do some research.

wilianmaique:
> try setting sv_send_logos 0 and sv_allowupload 0 and if you're using fast download sv_allowdownload 0. there's also sv_allow_dlfile too but idk what that does exactly. do some research.

not resolve

di57inct:
have you tried setting sv_allow_dlfile 0 too?
have you messed with any of these cvars?:
[...]

PR #1183, opened 2026-06-17
engine: add sv_reconnect_timeout to drop clients that never reconnect after a level change

brokuka:
Adds a server-side deadline for clients to re-initiate their connection after a level change. A client that the engine put into the reconnect-pending state on `changelevel` but that never sends a new connection request within `sv_reconnect_timeout` seconds is now force-dropped by `SV_CheckTimeouts`, independent of the netchan inactivity timeout.

This closes a phantom-slot vector that survives `changelevel` and is currently abusable as a lag/CPU (L7) amplifier. Related reports: #1098, #1054.

On `changelevel`, `SV_InactivateClients()` does **not** drop real clients — it flips them to `connected = TRUE, spawned = FALSE, fully_connected = FALSE` and the engine sends each one a `reconnect` command (`svc_stufftext`, see `SV_BuildReconnect`). The client is then expected to reconnect **on its own**.

If a client silently drops that `reconnect` command (a publicly available cheat does exactly this with an `svc_stufftext` filter) while keeping its netchan warm, the slot stays half-alive forever:

- `SV_CheckTimeouts()` is the only thing that could remove it, but it keys off `cl->netchan.last_received < realtime - sv_timeout`. A warm netchan keeps `last_received` fresh, so the inactivity timeout **never fires**.
- There is no separate "you must finish reconnecting within N seconds" deadline.

Result: a `connected && !spawned && !fully_connected` client persists across the map change. It occupies a slot, is counted as active, shows no ping on the scoreboard (`SV_CalcPing` has no frame samples for it), and — because it bypasses the spawned-only rate/flood paths — can be used to load the server / spike everyone's ping. Restarting the map clears it, which is why these reports look like an "exploit, not a DDoS".

`SV_InactivateClients()` arms no completion deadline, and `SV_CheckTimeouts()` measures liveness from `netchan.last_received`, which a client fully controls by sending any packet.

The fix tracks, per slot, the moment a client was inactivated by a level change, and enforces a hard deadline that is **independent of `last_received`**:

- `SV_InactivateClients()` — record `realtime` for each real client it inactivates (arm).
- `SV_ConnectClient_internal()` — clear it the moment the client re-establishes its connection (proof it obeyed `reconnect`). Cleared **before** any resource download/spawn, so downloading players are never affected.
- `SV_CheckTimeouts()` — if a slot is still armed and not `fully_connected` after `sv_reconnect_timeout` seconds, force-drop it.

The deadline only covers the narrow window "told to reconnect → has not yet sent a connection request". For any real client this is sub-second; only a client that never reconnects at all trips it.

State is kept in a file-local `static double[MAX_CLIENTS]` rather than a `client_t` field, to avoid changing the struct layout (`structSizeCheck`). Guarded by `REHLDS_FIXES`.

New cvar:

| Cvar | Default | Description |
|------|---------|-------------|
| `sv_reconnect_timeout` | `30` | Seconds a client may take to reconnect after a level change before being force-dropped. `0` disables the check. |

It only needs to cover the network handshake (not downloads), so the default is generous.

Reproduction:

1. Connect a client that filters the server `reconnect` stufftext command and keeps its netchan alive.
2. Trigger a `changelevel`.
3. Before: the client stays as a half-connected, ping-less phantom slot indefinitely; map restart is the only way to clear it.
4. After (with `sv_reconnect_timeout > 0`): the slot is dropped `sv_reconnect_timeout` seconds after the level change with reason `Failed to reconnect after level change`.

Testing:

- Built `Release|Win32` (MSVC), no struct-size assertion failures.
- Legit clients (including ones that download the new map via fastdl after reconnecting) are unaffected: their deadline is cleared in `SV_ConnectClient_internal` before downloads begin.
- Phantom slot is dropped at the deadline; server load / ping return to normal.

Notes:

- All changes are additive and gated behind `REHLDS_FIXES`.
- No `client_t` layout change.
- Backwards compatible: `sv_reconnect_timeout 0` restores the previous behavior exactly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,7 @@ This means that plugins that do binary code analysis (Orpheu for example) probab
 <li>sv_rehlds_movecmdtime_min_scale // Defines the min client's base game speed ratio. Clients slowing the game down below this multiplier will receive warnings. Default: 0.5
 <li>sv_rehlds_movecmdtime_max_warnings // Maximum allowed speedhack/slowmo warnings before the punishment is applied. -1 - disable detection. Default: -1
 <li>sv_rehlds_movecmdtime_punish // Time in minutes for which the player will be banned for speedhacking/slowing (-1 - Kick, 0 - Permanent, use a negative number for a kick). Default: -1
+<li>sv_reconnect_timeout // Hard deadline in seconds for a client to re-initiate its connection after a level change, independent of netchan activity. Closes a phantom-slot exploit where a cheat blocks the "reconnect" command and keeps the netchan warm so sv_timeout never fires. 0 - disabled. Default: 30
 </ul>
 </details>
 
--- a/rehlds/engine/sv_main.cpp
+++ b/rehlds/engine/sv_main.cpp
@@ -212,7 +212,6 @@ cvar_t sv_timeout = { "sv_timeout", "60", 0, 0.0f, NULL };
 // on changelevel and keeps the netchan warm, so the normal sv_timeout (which keys off
 // netchan.last_received) never fires and the half-connected slot survives forever. 0 = disabled.
 cvar_t sv_reconnect_timeout = { "sv_reconnect_timeout", "30", 0, 0.0f, NULL };
-static double g_ReconnectDeadline[MAX_CLIENTS];
 #endif // REHLDS_FIXES
 cvar_t sv_failuretime = { "sv_failuretime", "0.5", 0, 0.0f, NULL };
 cvar_t sv_cheats = { "sv_cheats", "0", FCVAR_SERVER, 0.0f, NULL };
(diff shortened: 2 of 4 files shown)


dystopm:
Nice catch. Took a look at your code. Think about extending CGameClient's class structure (instead of using global arrays), since it is created once in Rehlds_Interfaces_InitClients for each available slot. It would look cleaner.

brokuka:
Good point — moved the state into `CGameClient` (`m_reconnectDeadline` + `Get/SetReconnectDeadline`, gated by `REHLDS_FIXES`), following the existing `m_localGameTimeBase` / `m_bSpawnedOnce` pattern. Dropped the file-local `static double[MAX_CLIENTS]` array entirely; the three sites (`SV_InactivateClients` arm, `SV_ConnectClient_internal` clear, `SV_CheckTimeouts` check) now go through `g_GameClients[i]`, which `sv_main.cpp` already uses right next to these spots. Reads cleaner and the lifetime is tied to the slot object. Thanks!

RauliTop:
Please, add the cvar on the README.md

